PR #5, opened 2017-12-17
Use the cryptography package rather than PyOpenSSL

rgov:
Sorry, at the moment I cannot run the tests, but this should be a good starting point if anyone wants to clean it up.

- Removes PyOpenSSL as a runtime requirement. It is still a testing requirement.
- Changes serialization and deserialization methods to use `cryptography.x509` objects instead of the corresponding objects in `OpenSSL.crypto`.
- Extends the `ComparableX509` class to wrap `cryptography.x509` objects and to convert `OpenSSL.crypto` objects for compatibility.

Addresses #1.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,8 +10,6 @@ install_requires = [
     # load_pem_private/public_key (>=0.6)
     # rsa_recover_prime_factors (>=0.8)
     'cryptography>=0.8',
-    # Connection.set_tlsext_host_name (>=0.13)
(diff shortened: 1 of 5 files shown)


rgov:
Git how do you even work

rgov:
I don't strongly agree with the patch suggested by `isort`, but your call.

bmw:
I'm going to close this PR for now, but we're open to another if my review comments are kept in mind.

PR #7, opened 2018-01-18
Remove redundant Python 2.6 and 3.3 code

hugovk:
From .travis.yml and tox, it looks like only Python 2.7 and 3.4-3.6 are supported.

That means we can remove some redundant code for those EOL versions.


bmw:
Thanks for the PR @hugovk and I'm sorry it took me so long to get to it.

PR #8, opened 2018-01-22
Add support for Elliptic Curve keys

yonran:
Hi. I wanted to add EC support to certbot because I was using [lego](https://github.com/xenolf/lego/) and wanted to switch my account to certbot. But lego created an account signed using EC instead of RSA so my account was stuck. I would like guidance on what tests I should write and how to run pylint/pydoc.

### DIFF
--- a/src/josepy/jwk.py
+++ b/src/josepy/jwk.py
@@ -9,7 +9,7 @@ import six
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes  # type: ignore
 from cryptography.hazmat.primitives import serialization
-from cryptography.hazmat.primitives.asymmetric import ec  # type: ignore
+from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.hazmat.primitives.asymmetric import rsa
 
 from josepy import errors, json_util, util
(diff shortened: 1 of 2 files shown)


bmw:
Hey @yonran. Sorry this has sat around so long.

I would largely base the tests on the tests for the RSA versions of the classes you're adding. You can find these in `jwk_test.py` and `util_test.py`.

As for how to run tests, you can install josepy and it's dependencies in a virtual environment by running commands like the following from the root of the repo:
```
virtualenv venv
source venv/bin/activate
pip install -e .[dev,tests]
```
After that, you can run tests with `tox` by running:
```
tox -e py27
```
You can replace 27 whichever Python version you want to test against as long as that version of Python is installed on your system.

jannispinter:
I would love to see this and #45  implemented, as ES256 is a MUST requirement for ACME.

Quoting [RFC 8555 Section 6.2](https://tools.ietf.org/html/rfc8555#section-6.2) here:

> An ACME server MUST implement the "ES256" signature algorithm
   [RFC7518] and SHOULD implement the "EdDSA" signature algorithm using
   the "Ed25519" variant (indicated by "crv") [RFC8037].

bmw:
Indeed it is a MUST for an ACME server.

While the small Certbot team would also love to see this implemented, we probably won't be able to get to it for a little while. If you'd like to see it sooner, we'd accept well written PRs for these features.

bmw:
Now that #51 has landed, can this be closed?

cc @adferrand

adferrand:
Indeed! This PR was in fact the base on which the actually merged PR has been constructed.

JamesTheAwesomeDude:
Just to confirm: this only added the `SECP-` keys, and not `X25519`?

adferrand:
Exactly. A work is in progress to add the other elliptic curve keys: https://github.com/certbot/josepy/pull/98

PR #16, opened 2018-03-28
Use cryptography's sign/verify methods instead of signer/verifier

### LINKED ISSUE #13: Deprecation warning when signing/verifying JWKRSA objects
The cryptography extension has deprecated [1] (since 2.0 - 2017-07-17) the `signer` and `verifiier` methods, which results in this warning:
```
/var/www/workaround/venv/lib/python3.5/site-packages/josepy/jwa.py:107: CryptographyDeprecationWarning: signer and verifier have been deprecated. Please use sign and verify instead.
  signer = key.signer(self.padding, self.hash)
```

Code I used to get this warning:
```
from acme import client
from cryptography.hazmat.backends import default_backend
from cryptography.hazmat.primitives.asymmetric import rsa

import josepy as jose

private_key = rsa.generate_private_key(public_exponent=65537, key_size=4096, backend=default_backend())
acme_client = client.Client( 'https://acme-v01.api.letsencrypt.org/directory', jose.JWKRSA(key=private_key))
regr = acme_client.register()
```

Environment:
* acme-0.22.2
* josepy-1.0.1
* cryptography-2.2.1
* Python 3.5.3

[1] https://github.com/pyca/cryptography/issues/3659

mgedmin:
See https://github.com/pyca/cryptography/issues/3659.

Fixes #13.

P.S. I hope your test coverage is good because I've no idea how to test this
code manually!

### DIFF
--- a/src/josepy/jwa.py
+++ b/src/josepy/jwa.py
@@ -104,26 +104,18 @@ class _JWARSA(object):
     def sign(self, key, msg):
         """Sign the ``msg`` using ``key``."""
         try:
-            signer = key.signer(self.padding, self.hash)
+            return key.sign(msg, self.padding, self.hash)
         except AttributeError as error:
             logger.debug(error, exc_info=True)
             raise errors.Error("Public key cannot be used for signing")
         except ValueError as error:  # digest too large
             logger.debug(error, exc_info=True)
             raise errors.Error(str(error))
-        signer.update(msg)
-        try:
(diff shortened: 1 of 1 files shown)


bmw:
We need to keep backwards compatibility with older versions of `cryptography` so we went with the approach in #18, but I appreciate the PR and the ping about the issue. We'll be doing a release of `josepy` fixing this shortly.

PR #17, opened 2018-03-28
Use README.rst as long_description for PyPI

mgedmin:
Currently https://pypi.org/project/josepy/ says

  Project Description

  UNKNOWN

which is a bit unfriendly.  Let's fix that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,6 +40,10 @@ docs_extras = [
 ]
 
 
+with open('README.rst') as f:
+    long_description = f.read()
+
+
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -40,6 +40,10 @@
 ]
 
 
+with open('README.rst') as f:
```
sydneyli: nit: might as well use `io.open` with `utf-8` set?

mgedmin: Sure!

PR #32, opened 2018-05-17
Sets up mypy in tox

### LINKED ISSUE #29: Set up mypy tests in tox
This should be somewhat based on Certbot's `tox.ini`, but will be a lot simpler because we don't have wrappers around pip and our tests here. You'll probably also have to add a `mypy.ini` to the root of the repo which again, should be based on the file in Certbot. You shouldn't add any `check_untyped_defs` for now.

You should make sure these tests pass and fix any errors that prevent them from doing so. After they're passing, `.travis.yml` should be configured to run them.

jepayne1138:
Please carefully review the last commit made to make mypy tests pass (see commit message for more details)

Resolves #29

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,7 @@ var/
 pip-log.txt
 pip-delete-this-directory.txt
 
-# Unit test / coverage reports
+# Unit test / coverage / mypy reports
 htmlcov/
 .tox/
 .coverage
@@ -47,6 +47,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+.mypy_cache
 
 # Translations
 *.mo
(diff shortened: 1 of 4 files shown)


bmw:
@joohoi or @sydneyli, want to take a crack at reviewing this?

As stated above, the most recent commit should be reviewed somewhat carefully. It's probably fine, but we should spend sometime thinking and testing about this to make sure.

atombrella:
I continued this in #92

atombrella:
This PR can be closed https://github.com/certbot/josepy/commit/ef144a4a6aa6eac942f954eb2ea86108ff961a07 @jepayne1138 Thanks for the initial work!

PR #36, opened 2018-07-06
List py37 support and add tests

bmw:


bmw:
> I don't have the write privs so apparently it doesn't count for Github

You do now! You not having it already was just an oversight. Sorry about that!

PR #39, opened 2018-10-24
Ignore the new W504 error from flake8 to get the tests passing

alex:


ohemorange:
Not sure why this isn't working. Could try flipping the order of the ignores to see if that helps, but honestly it's only yelling about a few places so it might just be worth it to fix those.

alex:
Switched order doesn't seem to affect it.

Maybe the E501s can be fixed, but unfortunately I don't think the W504s can. W504 complains about a binary operator before a linebreak, there's also a warning about binary operators _after_ a linebreak, so you need to ignore at least one of them!

ohemorange:
This seems relevant: https://github.com/tholo/pytest-flake8/issues/8#issuecomment-270580940

pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
 addopts = -v --flake8 --isort --cov-report xml --cov-report=term-missing --cov=josepy --cov-config .coveragerc
 norecursedirs = *.egg .eggs dist build docs .tox
-flake8-ignore = E501
+flake8-ignore = W504,E501
```
ohemorange:
```suggestion
flake8-ignore = W504 E501
```

tox.ini
```diff
@@ -11,4 +11,4 @@ deps:
     .[tests]
 
 [flake8]
-ignore = E501
+ignore = W504,E501
```
ohemorange:
```suggestion
ignore = W504, E501
```

alex:
Nice catch! (Also the "suggest changes" UI is slick!)

PR #40, opened 2018-12-13
Make code compatible with Python >= 3.8 towards collections module

adferrand:
Starting with Python 3.8, abstract types from `collections` will need to be imported from the `collections.abc`. Currently on recent versions of Python, theses types can be imported both from `collections` and `collections.abc`, but will generated a warning about support drop for Python 3.8 if `collections` is used.

This PR allows to use `collections.abc`, and falls back to `collections` if necessary.

### DIFF
--- a/src/josepy/interfaces.py
+++ b/src/josepy/interfaces.py
@@ -1,14 +1,15 @@
 """JOSE interfaces."""
 import abc
 import json
-try:
-    from collections.abc import Sequence, Mapping  # pylint: disable=import-error
-except ImportError:
-    from collections import Sequence, Mapping
 
 import six
 
 from josepy import errors, util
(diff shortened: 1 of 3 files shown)


adferrand:
cc @bmw

bmw:
Looks like isort is unhappy. I personally think what it suggests is fine so I'd just do that if I were you.

Also, do you have a link about this change in Python 3.8?

adferrand:
Yes, isort is confused because the import statements are in a try except block. Developers warn about that in the documentation. Not a big deal so, I will make it happy.

For the breaking change, here is the commit in cpython that added the warn: c66f9f8d3909f588c251957d499599a1680e2320

PR #41, opened 2019-01-10
Don't require sudo

bmw:
Now that Travis has real [Xenial support](https://docs.travis-ci.com/user/reference/xenial/) we don't need to use `sudo: true` which should speed up the build slightly.


bmw:
Closing and reopening to trigger tests on travis-ci.com.

PR #42, opened 2019-01-10
Always use xenial

bmw:
Now that Travis has real [Xenial support](https://docs.travis-ci.com/user/reference/xenial/), I figure we might as well use it for all environments to get later and greater versions of software and simplify the Travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,14 @@
 language: python
 sudo: false
 cache: pip
+dist: xenial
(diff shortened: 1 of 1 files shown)


bmw:
Closing and reopening to trigger tests on travis-ci.com.

adferrand:
I could not agree more, since trusty is about to be EOL. 

LGTM.

PR #47, opened 2019-03-08
Random change to trigger CI

alex:
We're seeing josepy failures in the pyca/cryptography test suite and I want to see if it's an upstream problem

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,7 @@ envlist =
 commands =
     py.test {posargs}
 deps:
-    # installs the test dependencies as specified in setup.py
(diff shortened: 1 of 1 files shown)


alex:
Looks like `isort` isn't happy.

bmw:
Ah sorry for the trouble. Looks like the tests started failing with the new version of `isort`.

The behavior I'm seeing seems a little strange so I'm just removing the isort checks for now in #48.

alex:
Thanks for jumping on this! I'll defer to you on what the right fix is.

PR #56, opened 2020-01-24
Deprecate python 3.4 support

bmw:
Part of https://github.com/certbot/josepy/issues/53.


bmw:
I tested this warning was successfully raised using Python 3.4 from EPEL on CentOS 6.

PR #62, opened 2020-04-16
Update release.sh to Python 3 and fix virtualenv

bmw:
Fixes https://github.com/certbot/josepy/issues/61 and applies the same fix to `virtualenv` commands as https://github.com/certbot/certbot/pull/7832.

@ohemorange, are you able to review this small PR when you get a minute?

### DIFF
--- a/tools/release.sh
+++ b/tools/release.sh
@@ -46,7 +46,7 @@ mv "dist.$version" "dist.$version.$(date +%s).bak" || true
 git tag --delete "$tag" || true
 
 tmpvenv=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


bmw:
By the way, I left the use of `virtualenv` inside of the first virtual environment we create to help get an updated version of pip, setuptools, and wheel in the second virtual environment we make and to not make unnecessary changes in this PR. `virtualenv` will use Python 3 because [it defaults to whatever version of Python it's currently running on](https://virtualenv.pypa.io/en/latest/user_guide.html#introduction) and we installed it in a virtual environment explicitly created using Python 3.

PR #64, opened 2020-05-12
Pin flake8/all Python dependencies in tests

### LINKED ISSUE #63: Pin flake8/all Python dependencies in tests
[Travis tests are failing](https://travis-ci.com/github/certbot/josepy/builds/165358441) due to https://github.com/tholo/pytest-flake8/issues/66.

We can fix this and avoid issues like this in the future by providing a constraints file to `pip` pinning our dependencies.

Unless that pytest-flake8 issue is resolved and a release is put out in the next few hours, I think we should do this today.

When we do, I think it'd be nice to also submit a PR to cryptography updating https://github.com/pyca/cryptography/blob/2f2f3d2e414a0167ae3a98b9b608904b2c76a35f/.travis/downstream.d/certbot-josepy.sh to use this constraints file.

@alex & @reaperhulk, I'll make sure the Travis failures here and on cryptography's repo are fixed today one way or another.

EDIT: If we add a constraints file and configure cryptography to use it, I think we should add comments to that file reminding us that cryptography is relying on it so we can help them if we move it.

ohemorange:
Fixes #63.

Took some finagling to get a working set of dependencies. Some I copied from Certbot, some are based on going back through releases to find a compatible set.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,21 +1,33 @@
-# Specifies Python package versions for development and testing.
-# To generate:
-# python3 -m venv venv3 && source venv3/bin/activate
+# Specifies Python package versions for testing.
+# To update:
+# python3 -m venv venv && source venv/bin/activate
 # python3 -m pip install -e .[dev,tests] && pip freeze
+# Examine the produced list and update invididual entries here.
 # If anything changes about this file's contents or location, we may
 # need to submit a PR to pyca/cryptography, since they rely on it.
 
 apipkg==1.5
+atomicwrites==1.4.0
 attrs==19.3.0
 cffi==1.14.0
+configparser==4.0.2
+contextlib2==0.6.0.post1
 coverage==5.1
 cryptography==2.9.2
+entrypoints==0.3
+enum34==1.1.10
 execnet==1.7.1
 flake8==3.7.9
+funcsigs==1.0.2
+functools32==3.2.3.post2
+importlib-metadata==1.6.0
+ipaddress==1.0.23
 mccabe==0.6.1
 mock==1.3.0
 more-itertools==5.0.0
 packaging==20.3
+pathlib2==2.3.5
+pbr==5.4.5
 pluggy==0.13.1
 py==1.8.1
 pycodestyle==2.5.0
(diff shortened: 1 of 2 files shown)


constraints.txt
```diff
@@ -0,0 +1,31 @@
+# Specifies Python package versions for development and testing.
+# To generate:
+# python3 -m venv venv3 && source venv3/bin/activate
+# python3 -m pip install -e .[dev,tests] && pip freeze
```
bmw, line 4: nit: It seems like `dev` extras weren't used here because things like `tox` and its dependencies aren't pinned.

I think it's fine if they're not pinned for now since the motivation behind this issue was pinning our test dependencies, but it'd be good to properly document how this file was created.
```suggestion
# python3 -m pip install -e .[tests] && pip freeze
```

ohemorange: Ah, I should have added a comment -- that's necessary to get things working locally for me, so I left it to make things easier when running locally in the future. My intent was that people would run these commands, look at versions of things, and cherry-pick updates as necessary. I intentionally removed the dev-only dependencies since it looked like testing doesn't need those (I think it's installing tox itself and such that makes it necessary for me, and since this is used after installing tox in the tests, that's not necessary). This file isn't the output of those commands directly at all, since some versions needed to be changed, and also after adding the other ones you suggest. I'll add a comment to this effect if you're cool with that?

ohemorange:
That's interesting, not sure how mine is working without those, they don't show up in my test output at all, nor are they listed in `.tox/py38/lib/python3.8/site-packages/`. I don't have them installed globally on my machine, nor in the josepy venv. Are you sure we actually need these packages, and they're not somehow artifacts of something else?

constraints.txt
bmw: Sure. I personally don't have too strong of an opinion as long as our instructions here are accurate.

bmw:
> Are you sure we actually need these packages, and they're not somehow artifacts of something else?

Pretty sure and with this file being used as constraints rather than requirements, I think we should err on the side of including too much rather than too little.

To look into one of the packages though, this PR pins `pytest==4.6.10` which [depends unconditionally on `atomicwrites>=1.0`](https://github.com/pytest-dev/pytest/blob/4.6.10/setup.py#L13) and isn't pinned in `constraints.txt`.

PR #66, opened 2020-07-10
added examples/pem_coversion.py

jvanasco:
added an example script, as suggested in #65, illustrating how to convert certbot keys to/from PEM


hews:
I was just gearing up to stumble through this for myself when I saw your example. Thanks!

PR #76, opened 2021-01-06
Deprecate support for Python 2.7

bmw:
This was based on https://github.com/certbot/josepy/pull/67.

Once this PR is merged and included in a release, we can drop Python 2.7 support here as well.


ohemorange (approved):
lgtm!

PR #81, opened 2021-02-02
Remove six dependency

bmw:
Part of https://github.com/certbot/certbot/issues/8494.

We should keep `six` pinned because its still a transitive dependency of ours.


alexzorin (approved):
LGTM!

PR #82, opened 2021-02-02
Remove mock dependency

bmw:
Now that we're not supporting Python 2, we can just use the standard library.


alexzorin (changes requested):
Merge conflicts

bmw:
Thanks for taking a look. I fixed the merge conflicts.

alexzorin (approved):

PR #85, opened 2021-02-14
Updated unittest assert to Python 3 support.

atombrella:
They provide better output and debugging. I split also a couple of instances of `assertTrue(all` to call `assertIs` for each of the statements.

### DIFF
--- a/src/josepy/interfaces_test.py
+++ b/src/josepy/interfaces_test.py
@@ -66,7 +66,7 @@ class JSONDeSerializableTest(unittest.TestCase):
 
     def test_to_json_other(self):
         mock_value = object()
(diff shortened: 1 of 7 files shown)


bmw:
Thanks for yet another PR! It looks like there are some minor merge conflicts here from some of our other cleanup work from dropping Python 2 support. Do you have the time and interest to resolve them?

atombrella:
@bmw It should be updated, and passing locally :)

https://docs.python.org/3/library/unittest.html?highlight=assertin#unittest.TestCase.assertTrue also has advice about `assertTrue`/`assertFalse`.

bmw (approved):
Thanks again :)

PR #86, opened 2021-02-14
collections.abc is supported on all supported Python versions.

atombrella:
Should be good as of Python 3.3+ :) Will also increase code coverage slightly by removing some dead catch-blocks ;)


alexzorin (approved):
LGTM, thanks!

PR #88, opened 2021-02-26
Deprecate magic_typing

bmw:
Fixes https://github.com/certbot/certbot/issues/8660.

PR #90, opened 2021-02-26
Remove broken test for typing import failure

bmw:
This test doesn't work with newer versions of `cryptography`. See https://github.com/certbot/josepy/pull/88#issuecomment-786911443.

Since `magic_typing` is deprecated and `typing` should always be available since we only support Python 3.6+, I think we should just delete this test.


ohemorange (approved):
The different import order stuff in different unit testing methods is uncomfortable, but I agree that despite that this test is no longer necessary.

PR #91, opened 2021-02-27
EC signature padding

commonism:
The r||s encoding is wrong for ec keys and signatures with leading zeros.
This results in bad signatures, I've come across:

"Parse error reading JWS: EC public key has incorrect padding"
"failed to unmarshal JWK: square/go-jose: invalid EC public key, wrong length"

Leading zero happens very frequently for ES512, rare with ES256.

This was already wrong in my proposal for r||s and just got ported to using cryptography api.

int.to_bytes requires python >= 3.2
bytes.fromhex python >= 3.0
so they are safe to use here.

### DIFF
--- a/src/josepy/jwa.py
+++ b/src/josepy/jwa.py
@@ -172,8 +172,9 @@ class _JWAEC(JWASignature):
         """Sign the ``msg`` using ``key``."""
         sig = self._sign(key, msg)
         dr, ds = decode_dss_signature(sig)
-        return (dr.to_bytes(length=math.ceil(dr.bit_length() / 8), byteorder='big') +
-                ds.to_bytes(length=math.ceil(ds.bit_length() / 8), byteorder='big'))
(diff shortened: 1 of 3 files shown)


commonism:
Had some issues with the CI, force pushed to keep things tidy.

bmw:
@atombrella or @adferrand, do either of you have the time and interest to take a look at this since you reviewed https://github.com/certbot/josepy/pull/51?

adferrand:
For sure I will review it in the next days. I already looked at it quickly this week and it seemed good.

bmw:
Thanks Adrien!

I'm assigning you for now but if you end up not being able to get to it for whatever reason, just let me know.

adferrand (approved):
Thanks a lot @commonism, this LGTM. I appreciate in particular the updated test cases that will catch this error if future developments would break again when leading zeros are involved, since this is something quite hard to catch without unit tests.

PR #92, opened 2021-03-02
Added mypy to tests

atombrella:
My experience with static type checking and MyPy is a bit limited, but it's always a good way to educate yourself :)

There are still a bunch `:rtype:` annotations left, which I propose to replace with PEP484 syntax.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,13 @@ dist: xenial
 matrix:
   include:
     - python: "3.6"
-      env: TOXENV=py36, mypy
+      env: TOXENV=py36,mypy
     - python: "3.7"
-      env: TOXENV=py37, mypy
+      env: TOXENV=py37,mypy
     - python: "3.8"
-      env: TOXENV=py38, mypy
+      env: TOXENV=py38,mypy
     - python: "3.9"
-      env: TOXENV=py39, mypy
+      env: TOXENV=py39,mypy
 install:
   - pip install tox-travis codecov
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ matrix:
     - python: "3.7"
     - python: "3.8"
     - python: "3.9"
-    - mypy
 install:
   - pip install tox-travis codecov
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ dist: xenial
 matrix:
   include:
     - python: "3.6"
+      env: TOXENV=py36, mypy
     - python: "3.7"
     - python: "3.8"
     - python: "3.9"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ matrix:
     - python: "3.7"
     - python: "3.8"
     - python: "3.9"
+    - mypy
 install:
   - pip install tox-travis codecov
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,11 @@ matrix:
     - python: "3.6"
       env: TOXENV=py36, mypy
     - python: "3.7"
+      env: TOXENV=py37, mypy
     - python: "3.8"
+      env: TOXENV=py38, mypy
     - python: "3.9"
+      env: TOXENV=py39, mypy
 install:
   - pip install tox-travis codecov
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 sudo: false
 cache: pip
-dist: xenial
+dist: bionic
 matrix:
   include:
     - python: "3.6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,11 @@ dist: bionic
 matrix:
   include:
     - python: "3.6"
-      env: TOXENV=py36,mypy
     - python: "3.7"
-      env: TOXENV=py37,mypy
     - python: "3.8"
-      env: TOXENV=py38,mypy
     - python: "3.9"
-      env: TOXENV=py39,mypy
+    - python: "3.9"
+      env: TOXENV=mypy
 install:
   - pip install tox-travis codecov
 script:
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,3 @@
+[mypy]
+python_version = 3.6
+ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,3 +1,2 @@
 [mypy]
-python_version = 3.6
 ignore_missing_imports = True
--- a/setup.py
+++ b/setup.py
@@ -73,6 +73,7 @@ setup(
         'dev': dev_extras,
         'docs': docs_extras,
         'tests': testing_requires,
+        'mypy': ['mypy']
     },
     entry_points={
         'console_scripts': [
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,6 @@ setup(
         'dev': dev_extras,
         'docs': docs_extras,
         'tests': testing_requires,
-        'mypy': ['mypy']
     },
     entry_points={
         'console_scripts': [
--- a/src/josepy/json_util.py
+++ b/src/josepy/json_util.py
@@ -9,6 +9,7 @@ The framework presented here is somewhat based on `Go's "json" package`_
 import abc
 import binascii
 import logging
+from typing import Any, Dict
 
 import OpenSSL
 
@@ -159,6 +160,8 @@ class JSONObjectWithFieldsMeta(abc.ABCMeta):
 
     """
 
+    _fields: Dict[str, Field] = {}
+
     def __new__(mcs, name, bases, dikt):
         fields = {}
 
@@ -210,7 +213,7 @@ class JSONObjectWithFields(util.ImmutableMap,
     """
 
     @classmethod
-    def _defaults(cls):
+    def _defaults(cls) -> Dict:
         """Get default fields values."""
         return {
             slot: field.default for slot, field in cls._fields.items()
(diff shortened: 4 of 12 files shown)


src/josepy/json_util.py
```diff
@@ -409,17 +412,17 @@ def decode_csr(b64der):
 class TypedJSONObjectWithFields(JSONObjectWithFields):
     """JSON object with type."""
 
-    typ = NotImplemented
+    typ = NotImplemented  # type: Any
```
atombrella, line 415: Without this, there are quite a lot of warnings about `_NotImplementedType` in the subclasses.

src/josepy/json_util.py
```diff
@@ -157,6 +158,8 @@ class Foo(object):
 
     """
 
+    _fields = {}  # type: Dict[str, Any]
```
alexzorin, line 161: Since josepy requires Python 3.6+, I think we can replace this and all other instances of mypy `# type:` comments with PEP484 type hints, as you mentioned:

```suggestion
    _fields: Dict[str, Any] = {}
```

There's not that many in the project, maybe we can do it now?

src/josepy/jwk_test.py
```diff
@@ -27,7 +28,7 @@ def test_load_subclass_wrong_type(self):
 class JWKTestBaseMixin(object):
     """Mixin test for JWK subclass tests."""
 
-    thumbprint = NotImplemented
+    thumbprint = NotImplemented  # type: Any
```
alexzorin, line 31:
```suggestion
    thumbprint: bytes = NotImplemented
```

tox.ini
```diff
@@ -11,3 +11,11 @@ deps =
...
 [flake8]
 ignore = W504, E501
+
+[testenv:mypy]
+basepython = python3
+commands =
+    mypy src/josepy
+deps:
+    # installs the test dependencies as specified in setup.py
+    .[dev]
```
alexzorin, line 21: nit: to be consistent with the existing style of this file.

```suggestion
deps =
    -e .[dev]
```

src/josepy/jwk.py
```diff
@@ -16,15 +18,15 @@
...
     """JSON Web Key."""
     type_field_name = 'kty'
-    TYPES = {}  # type: dict
-    cryptography_key_types = ()  # type: tuple
+    TYPES = {}  # type: Dict
+    cryptography_key_types = ()  # type: Any
     """Subclasses should override."""
 
-    required = NotImplemented
+    required = NotImplemented  # type: Any
```
alexzorin, line 29:
```suggestion
    required: Sequence[str] = NotImplemented 
```

src/josepy/json_util.py
alexzorin, line 415: Reading over https://github.com/python/mypy/issues/4791, it seems like the best choice here would be:

```suggestion
    typ: str = NotImplemented 
```

src/josepy/jwk.py
```diff
@@ -16,15 +18,15 @@
...
 
-class JWK(json_util.TypedJSONObjectWithFields):
+class JWK(json_util.TypedJSONObjectWithFields, metaclass=abc.ABCMeta):
     # pylint: disable=too-few-public-methods
     """JSON Web Key."""
     type_field_name = 'kty'
-    TYPES = {}  # type: dict
-    cryptography_key_types = ()  # type: tuple
+    TYPES = {}  # type: Dict
+    cryptography_key_types = ()  # type: Any
```
alexzorin, line 26: Perhaps:

```suggestion
    cryptography_key_types: Sequence[Type] = () 
```

src/josepy/jwk.py
```diff
@@ -16,15 +18,15 @@
...
 
 
-class JWK(json_util.TypedJSONObjectWithFields):
+class JWK(json_util.TypedJSONObjectWithFields, metaclass=abc.ABCMeta):
     # pylint: disable=too-few-public-methods
     """JSON Web Key."""
     type_field_name = 'kty'
-    TYPES = {}  # type: dict
-    cryptography_key_types = ()  # type: tuple
+    TYPES = {}  # type: Dict
```
alexzorin, line 25: Is it this?

```suggestion
    TYPES: Dict[str, Type] = {}
```

src/josepy/json_util.py
```diff
@@ -409,17 +412,17 @@ def decode_csr(b64der):
...
-    type_field_name = "type"
+    type_field_name = "type"  # type: str
     """Field name used to distinguish different object types.
 
     Subclasses will probably have to override this.
 
     """
 
-    TYPES = NotImplemented
+    TYPES = NotImplemented  # type: Any
```
alexzorin, line 425:
```suggestion
    TYPES: Dict[str, Type] = NotImplemented
```

src/josepy/json_util.py
alexzorin, line 161:
```suggestion
    _fields: Dict[str, Field] = {}
```

alexzorin (changes requested):
Thanks!

I think we should aim to avoid `Any` as much as possible. From what I can see, there's only one place where we might need to keep it.

This feels like a good time to migrate to PEP484 Type Hints as well.

src/josepy/json_util.py
atombrella: Yes. I'll run https://github.com/ilevkivskyi/com2ann on the project :)

src/josepy/jwk.py
atombrella: I read through the documentation of Typing, and I don't see `Type` in the official documentation. It makes me a bit uncomfortable going with this suggestion (I committed the change, but it can always be handled locally). I'll search for an alternative that's documented.

bmw:
It looks like tests are failing. @atombrella, do you have the time and interest to investigate the issue and fix it?

atombrella:
@bmw Yes! I was just busy with other things the past week or so. The failing tests are primarily due to missing imports (I committed the suggestions by Alex directly).

bmw:
Thanks! There's no rush. I just commented to make sure that you and Alex were on the same page about this PR and you both didn't think you were waiting on each other to do something.

ohemorange:
Hey, just checking in on this! If you're hitting any weird issues with mypy, please don't hesitate to ask -- having spent some time on this in the past, I know getting it all clear can get tricky.

atombrella:
Sorry for the silence/inactivity. I'll push an update during Easter, so we can get this merged. I've been somewhat busy with work-stuff. My only "concern" about this PR is that `Type` is not in the official documentation for `typing`, so it feels less right to use it than `Any` which might be a wrong way of approaching `typing`.

alexzorin:
>My only "concern" about this PR is that Type is not in the official documentation for typing

I found it on documentation here: https://docs.python.org/3.6/library/typing.html#typing.Type. I do hope we can use it :D.

atombrella:
@alexzorin I've updated this. I found a few `type: ignore` which probably can be deleted (the commit can be deleted/reverted if it's a bit out of scope for this PR).

ohemorange (changes requested):
Hey, Alex is out, so I took a look at this -- looks like all the previous concerns were addressed, and what I see looks good to me. Kind of surprised removing the type ignores just works, but the test seems to pass so that's fine for me. Speaking of which, can we add the `mypy` test to the travis matrix in `.travis.yml` so it runs for the PR automatically? Other than that, looking good!

atombrella:
I'll add mypy to the matrix, and then we're good to go, I think.

mypy.ini
```diff
@@ -0,0 +1,3 @@
+[mypy]
+python_version = 3.6
```
ohemorange, line 2: Aw hey just noticed this line while doing a final check. Do we need it? We removed it from the analogous certbot file in https://github.com/certbot/certbot/pull/8426.

ohemorange:
Hm, looks like the mypy test still didn't run? https://travis-ci.com/github/certbot/josepy/builds/223307481

tox.ini
```diff
@@ -12,7 +12,7 @@ deps =
 [flake8]
 ignore = W504, E501
 
-[testenv:mypy]
+[mypy]
```
ohemorange, line 15:
```suggestion
[tentenv:mypy]
```

I think it does in fact need the `testenv`

atombrella: Yes. I'm fumbling a bit to figure out why `tox -v` runs mypy locally, but not here in Travis. I the change.

ohemorange:
As you may have already found, here's how we used to do it in the main certbot repo: https://github.com/certbot/certbot/blob/2af297d72ff21ea996e8972f9bc4a44f16b4a876/.travis.yml

atombrella:
@ohemorange Yes, I think it's fine now :)

atombrella:
I think commits should be squashed (preferably without the trial-and-error commit messages) :) Please note the "Co-Authored-By" when squashing; it needs to be the last line in the commit message in order for GitHub to display both authors.

setup.py
```diff
@@ -73,6 +73,7 @@
         'dev': dev_extras,
         'docs': docs_extras,
         'tests': testing_requires,
+        'mypy': ['mypy']
```
ohemorange, line 76: Presumably this should be under `dev_extras` so it's included in the `deps = -e .[dev]` call; is this called somewhere else that I'm not seeing?

.travis.yml
```diff
@@ -1,13 +1,17 @@
 language: python
 sudo: false
 cache: pip
-dist: xenial
+dist: bionic
 matrix:
   include:
     - python: "3.6"
+      env: TOXENV=py36,mypy
```
ohemorange, line 8: Glad to see we got the syntax down! That being said, I don't think we actually need to run mypy four times; shouldn't once be enough? And could we put it in its own toxenv so it's the only thing being called for that test run? This should look like adding, for example:


```
    - python: "3.9"
      env: TOXENV=mypy
```

ohemorange (changes requested):

atombrella:
I think commits should be squashed (preferably without the trial-and-error commit messages) :) Please note the "Co-Authored-By" when squashing; it needs to be the last line in the commit message in order for GitHub to display both authors.

.travis.yml
atombrella: OK :+1:

ohemorange (approved):
LGTM! Thanks!

PR #95, opened 2021-03-16
JWKEC - key en/de-coding

commonism:
Hi,

this patch changes the encoding of ec coordinate parameters to be static in length according to rfc.

Without this patch there is a chance to run into
```
Deserialization error: Expected parameter "{x,y}" to be range(a, b) bytes after base64-decoding; got c bytes instead
```


adferrand (approved):
For the record, the related RFC about fixed `x` and `y` sizes: https://tools.ietf.org/html/rfc7518#section-6.2.1.2.

LGTM!

PR #97, opened 2021-05-08
Python 3

atombrella:
Let's remove a few more remnants from the Python 2 days :)


atombrella:
f-strings could be another neat thing to introduce. They're somewhat faster than `format`.
https://realpython.com/python-f-strings/#speed

alexzorin (approved):
LGTM

PR #100, opened 2021-06-09
Update constraints and pin mypy

bmw:
These constraints hadn't been updated since we added mypy to our dev or tests extras so mypy wasn't pinned and a new release came out yesterday causing our tests to fail. This PR adds mypy to constraints and pins it to the previously working version for now.

Additionally, the constraints file wasn't being used in the tox config for mypy. To fix this, I moved mypy to the test extras and deleted the mypy deps setting in the tox config so it is inherited from the default testenv which uses the constraints file.

I also added quotes around .[dev,tests] because without it zsh complains:
```
$ python3 -m pip install -e .[dev,tests] && pip freeze
zsh: no matches found: .[dev,tests]
```

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -16,6 +16,7 @@ distlib==0.3.2
 execnet==1.8.1
 filelock==3.0.12
 flake8==3.9.2
+importlib-metadata==4.4.0
 iniconfig==1.1.1
 mccabe==0.6.1
 mypy==0.812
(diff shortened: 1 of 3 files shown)


bmw:
Tests are finally running. They broke on old versions of Python, but I fixed it and created https://github.com/certbot/josepy/issues/101 describing the problem.

ohemorange (approved):
lgtm

PR #104, opened 2021-07-10
Replace TravisCI with GitHub workflow

atombrella:
I propose to replace Travis with GitHub Workflow. It integrates better with GitHub, supports secrets under the settings, so the `token` for irc could be posted there.

I wonder why things somehow are directed via another tool to Mattermost.

The workflow file can be tweaked and cleaned up.


zoracon (commented):
Adding in some amendments. But overall looks good. Waiting on the builds and the status now. Will have to do some clean up when we switch from travis and add new badges to the repo.

Thank you!

adferrand:
I will just add one comment, since I carried out the migration from Travis CI to Azure Pipeline for the certbot project.

For you information, GitHub Actions is in fact a GitHub flavor of Azure Pipelines. So the two systems are very close together, with some features not (yet) implemented in GitHub Actions.

This is because of these limitations and the complexity of the certbot pipeline that we decided to go for Azure, but with a long term goal of moving to GitHub once the service has improved. I do regularly a review of GitHub Actions for that purpose.

However for josepy I think that GitHub Actions is totally sufficient for its context, and this will give us a first concrete implementation that would help assessing its viability for certbot.

zoracon (approved):

PR #105, opened 2021-07-19
Amend Travis Notifications for Mattermost Webhook

zoracon:
- Moving from IRC bridge and integrating directly into Mattermost
- review note: it seems like the travis-cli "cleaned up" the file when I added the webhook.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,11 +21,7 @@ notifications:
   webhooks:
     if: branch = master
     urls:
-    # This is set to a secure variable to prevent forks from sending
-    # notifications. This value was created by installing
-    # https://github.com/travis-ci/travis.rb and running
-    # `travis encrypt https://opensource.eff.org/hooks/...`.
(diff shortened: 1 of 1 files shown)


zoracon:
Webhook is not firing, so please hold on review.

bmw:
I think the main problem here is Travis' webhook format is different than Mattermost expects. Here's [the documentation for Travis' webhooks](https://docs.travis-ci.com/user/notifications/#configuring-webhook-notifications) and here is the [documentation for Mattermost's](https://developers.mattermost.com/integrate/incoming-webhooks/). Notice the lack of things like the `text` field which Matttermost expects in the example Travis webhook payload. [I've tried to get Travis to support Mattermost before](https://travis-ci.community/t/feature-request-mattermost-notifications/2992), even offering to write a PR in the issue linked in that post but I never got any response.

I think our best bet to configure Mattermost notifications is to switch CI to Azure Pipelines or GitHub Actions. Either works just fine, but I'd personally prefer the latter as I suspect it'll be a better option long term and [Mattermost maintains an action for it](https://github.com/marketplace/actions/mattermost). For Azure Pipelines on the Certbot repo, we're using our own custom code which you can see [here](https://github.com/certbot/certbot/blob/10eecf9c978ce909f6965de5d82cea39a720facc/.azure-pipelines/templates/stages/notify-failure-stage.yml#L12-L18). We also recently got a PR to move this repo to GitHub actions at https://github.com/certbot/josepy/pull/104.

(Finally, I think the reason Travis isn't running on your latest commit is that if you specify `on_success`, `on_failure`, etc. you need to put the syntax:
```
webhook:
  urls:
    - <url>
```
Travis' documentation isn't super clear on this, but using this syntax seems to work and you can see Travis complaining about parsing the config [here](https://travis-ci.com/github/certbot/josepy/requests).)

zoracon:
Going to look into Github Actions and close this out.

PR #112, opened 2021-09-12
Add types to functions (out of tests modules) in josepy

adferrand:
As part of a global effort to bring more types in the Certbot project, this PR adds type annotations to all methods/functions in `josepy` out of tests modules.

### DIFF
--- a/src/josepy/jws.py
+++ b/src/josepy/jws.py
@@ -373,7 +373,7 @@ class CLI:
             print(sig.json_dumps_pretty())
 
     @classmethod
-    def verify(cls, args: argparse.Namespace) -> int:
+    def verify(cls, args: argparse.Namespace) -> bool:
         """Verify."""
         if args.compact:
             sig = JWS.from_compact(sys.stdin.read().encode())
@@ -382,7 +382,7 @@ class CLI:
                 sig = cast(JWS, JWS.json_loads(sys.stdin.read()))
             except errors.Error as error:
                 print(error)
-                return -1
+                return False
 
         if args.key is not None:
             assert args.kty is not None
@@ -409,7 +409,7 @@ class CLI:
         return jwk.JWK.TYPES[arg]
 
     @classmethod
-    def run(cls, args: List[str] = None) -> Any:
+    def run(cls, args: List[str] = None) -> Optional[bool]:
         """Parse arguments and sign/verify."""
         if args is None:
             args = sys.argv[1:]
--- a/src/josepy/jws.py
+++ b/src/josepy/jws.py
@@ -73,7 +73,7 @@ class Header(json_util.JSONObjectWithFields):
     cty = json_util.Field('cty', encoder=MediaType.encode,
                           decoder=MediaType.decode, omitempty=True)
     crit = json_util.Field('crit', omitempty=True, default=())
-    _fields: Dict[str, Any]
+    _fields: Dict[str, json_util.Field]
 
     def not_omitted(self) -> Dict[str, Any]:
         """Fields that would not be omitted in the JSON object."""
--- a/src/josepy/jws.py
+++ b/src/josepy/jws.py
@@ -211,7 +211,7 @@ class Signature(json_util.JSONObjectWithFields):
         :param JWK key: Key for signature.
         :param JWASignature alg: Signature algorithm to use to sign.
         :param bool include_jwk: If True, insert the JWK inside the signature headers.
-        :param set protect: List of headers to protect.
+        :param FrozenSet protect: List of headers to protect.
 
         """
         assert isinstance(key, alg.kty)
--- a/src/josepy/jws.py
+++ b/src/josepy/jws.py
@@ -75,7 +75,7 @@ class Header(json_util.JSONObjectWithFields):
     crit = json_util.Field('crit', omitempty=True, default=())
     _fields: Dict[str, json_util.Field]
 
-    def not_omitted(self) -> Dict[str, Any]:
+    def not_omitted(self) -> Dict[str, json_util.Field]:
         """Fields that would not be omitted in the JSON object."""
         return {name: getattr(self, name)
                 for name, field in self._fields.items()
--- a/src/josepy/jws.py
+++ b/src/josepy/jws.py
@@ -2,9 +2,11 @@
 import argparse
 import base64
 import sys
+from typing import Dict, Any, Optional, FrozenSet, Mapping, List, Type, cast
 
-import OpenSSL
+from OpenSSL import crypto
 
+import josepy
 from josepy import b64, errors, json_util, jwa, jwk, util
 
 
@@ -15,7 +17,7 @@ class MediaType:
     """MIME Media Type and Content Type prefix."""
 
     @classmethod
-    def decode(cls, value):
+    def decode(cls, value: str) -> str:
         """Decoder."""
         # 4.1.10
         if '/' not in value:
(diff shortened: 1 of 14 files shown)


src/josepy/jws.py
```diff
@@ -181,27 +185,33 @@ def _with_combined(cls, kwargs):
...
+    def sign(cls, payload: bytes, key: josepy.JWK, alg: josepy.JWASignature,
+             include_jwk: bool = True, protect: FrozenSet = frozenset(),
+             **kwargs: Any) -> 'Signature':
         """Sign.
 
+        :param bytes payload: Payload to sign.
         :param JWK key: Key for signature.
+        :param JWASignature alg: Signature algorithm to use to sign.
+        :param bool include_jwk: If True, insert the JWK inside the signature headers.
+        :param set protect: List of headers to protect.
```
bmw, line 214:
```suggestion
        :param FrozenSet protect: List of headers to protect.
```

src/josepy/jws.py
```diff
@@ -252,21 +262,23 @@ class JWS(json_util.JSONObjectWithFields):
```
bmw, line 262: Can we update this incorrect docstring in this or another PR? You didn't change it, but you figured out what the types actually are here.

(On a larger level, I wonder if there's something we that can be done to reduce the duplication between our type annotations and the types in our docstrings.)

src/josepy/jws.py
```diff
@@ -361,13 +373,13 @@ def sign(cls, args):
             print(sig.json_dumps_pretty())
 
     @classmethod
-    def verify(cls, args):
+    def verify(cls, args: argparse.Namespace) -> int:
```
bmw, line 376: I think this should probably be:
```suggestion
    def verify(cls, args: argparse.Namespace) -> bool:
```
and the `return -1` below is converted to `return False`. What do you think?

src/josepy/jws.py
```diff
@@ -379,25 +391,25 @@ def verify(cls, args):
...
 
     @classmethod
-    def _kty_type(cls, arg):
+    def _kty_type(cls, arg: Any) -> Type[jwk.JWK]:
         assert arg in jwk.JWK.TYPES
         return jwk.JWK.TYPES[arg]
 
     @classmethod
-    def run(cls, args=None):
+    def run(cls, args: List[str] = None) -> Any:
```
bmw, line 412: Maybe
```suggestion
    def run(cls, args: List[str] = None) -> Optional[bool]
```
if the above suggestion is also taken?

I think we should probably rip out all of this code long term though.

src/josepy/util.py
```diff
@@ -1,31 +1,12 @@
...
+from typing import Union, Any, Callable, Iterator, Tuple
 from collections.abc import Hashable, Mapping
 
-import OpenSSL
+from OpenSSL import crypto
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import ec, rsa
 
 
-class abstractclassmethod(classmethod):
```
bmw, line 9: I think this is technically a part of the public API. See https://josepy.readthedocs.io/en/stable/api/util/#josepy.util.abstractclassmethod.

I agree we should deprecate and remove it at some point though.

src/josepy/util.py
```diff
@@ -54,22 +35,22 @@ def _dump(self, filetype=OpenSSL.crypto.FILETYPE_ASN1):
         :rtype: str
```
bmw, line 35:
```suggestion
        :rtype: bytes
```

src/josepy/util.py
```diff
@@ -99,12 +85,12 @@ def __eq__(self, other):
...
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return '<{0}({1!r})>'.format(self.__class__.__name__, self._wrapped)
 
-    def public_key(self):
+    def public_key(self) -> 'ComparableKey':
         """Get wrapped public key."""
-        return self.__class__(self._wrapped.public_key())
+        return self.__class__(self._wrapped.public_key())  # type: ignore[union-attr]
```
bmw, line 93: This is a bit of a nitpick, but I think this is a weird function here and in `ComparableECKey` below. I think `self._wrapped` will blow up if `self._wrapped` is either public key type right? If that's the behavior we want (which seems fine to me personally), why not check for a `public_key` attribute and error if it doesn't exist. Alternatively, if we want to support public keys like we try to in `ComparableECKey` maybe we could use the private numbers approach (or just return the public key directly)?

My hope is either approach would allow us to remove the `type: ignore` lines. What do you think?

bmw (changes requested):
I made a couple small things, but this largely LGTM!

As more and more of our code is annotated, I think we should play with `mypy` strictness options. One option is `mypy --strict`, however, I personally think we probably at least want to turn on [disallow_untyped_defs](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_untyped_defs) once the code is fully annotated. What do you think?

src/josepy/jws.py
adferrand: Yes here and in the `acme`  module, I just gave up editing every docstring that was either with wrong types arguments, missing arguments or invalid return, because it would increase the PR size without an immediate value (well, everyone is living with this documentation right ?).

So we can tackle the documentation issue in a further PR. During my study some potential toolings between the type annotations and the docstring, so I hope we will be able to fix most of the inconsistencies in an automated way.

src/josepy/jws.py
adferrand: Indeed!

src/josepy/jws.py
adferrand: Also done.

src/josepy/util.py
adferrand: Ok. I reimplemented the decorator `abstractclassmethod` using the standard decorators, and put the usual warning approach for deprecated attributes modules in place.

src/josepy/util.py
adferrand: Yes with `private_numbers` and self return in case of a public key, we can have a consistent behavior. I put a proposition in the code which allows to merge the `public_key()`  method in `ComparableKey`.

src/josepy/util.py
```diff
@@ -1,29 +1,19 @@
...
+import abc
+from types import ModuleType
+from typing import Union, Any, Callable, Iterator, Tuple, List, cast
 from collections.abc import Hashable, Mapping
+import sys
+import warnings
 
-import OpenSSL
+from OpenSSL import crypto
 from cryptography.hazmat.backends import default_backend
```
bmw, line 10: I think this is needed to make tests pass if you implement my other suggestion.
```suggestion
```

src/josepy/util.py
```diff
@@ -99,12 +94,17 @@ def __eq__(self, other):
...
-    def public_key(self):
+    def public_key(self) -> 'ComparableKey':
         """Get wrapped public key."""
-        return self.__class__(self._wrapped.public_key())
+        if isinstance(self._wrapped, (rsa.RSAPublicKeyWithSerialization,
+                                      ec.EllipticCurvePublicKeyWithSerialization)):
+            return self
+
+        key = self._wrapped.private_numbers().public_numbers.public_key(default_backend())
+        return self.__class__(key)
```
bmw, line 107: I think we can just do this right?
```suggestion
        return self.__class__(self._wrapped.public_key())
```

bmw (changes requested):
One nitpick/simplification, but this otherwise LGTM!

bmw (approved):

PR #113, opened 2021-09-14
Migrate tests to a separate folder and ensure all functions have type annotations

adferrand:
This PR is a follow-up to #112. It ensures that all functions in `josepy` source code are properly annotated.

In the process the tests modules are moved out of the source code, following the `src` layout that is strongly suggested in https://docs.pytest.org/en/6.2.x/goodpractices.html and described in https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure.


bmw (changes requested):
Thanks Adrien!

I have two suggestions:

1. Can we also move `src/josepy/testdata` out of `src/josepy`?
2. The unit tests are no longer being included in the built package. I think they should be and we should make sure doing something like building an sdist, installing it, and running `pytest --pyargs josepy` runs our unit tests.

adferrand:
For 1., sure !

For 2., the whole point of this layout is to not ship test code with the distribution. I think it is a sane approach as pytest and the article explain. Is there a specific reason to do so ? Otherwise I could as well find another solution specifically to ignore the tests modules in the type checks.

bmw:
I think (2) is valuable because OS packagers on distros like Debian, CentOS, Fedora, and Ubuntu run these tests to ensure everything is working properly. I think this layout has other benefits though and I believe the packages in the Certbot repo have behavior similar to what we both want. See https://github.com/certbot/certbot/blob/master/letstest/scripts/test_sdists.sh.

I think I was incorrect to include `--pyargs` in my suggested `pytest` command in my previous post though.

adferrand:
Yes I am aware of the test_sdist approach in certbot and the question can be extended to certbot itself, because it was my intention to propose the src layout to certbot projects also at some point.

I failed to understand from the source of the Debian package, but is it really how the packages are tested using the sdist, instead of the upstream source ?

bmw:
> I failed to understand from the source of the Debian package, but is it really how the packages are tested using the sdist, instead of the upstream source ?

Yes that's my understanding based on my previous conversations with Harlan, code like [this](https://salsa.debian.org/letsencrypt-team/certbot/certbot/-/commit/66f413a923366462271bdf519ecfdfbbc3af005a#68ef9f98c01c7eecd4c605cc26048a06f3304b79_3_3), and the fact the packages are split like our PyPI releases.

We can talk about this more if you want, but I want to flag that `pytest` itself has the `src` layout described above and includes their tests in their package like I'm proposing for. See:
```
$ wget https://files.pythonhosted.org/packages/4b/24/7d1f2d2537de114bdf1e6875115113ca80091520948d370c964b88070af2/pytest-6.2.5.tar.gz
$ tar xvf pytest-6.2.5.tar.gz
$ ls pytest-6.2.5
AUTHORS  CHANGELOG.rst  CITATION  CODE_OF_CONDUCT.md  CONTRIBUTING.rst  LICENSE  OPENCOLLECTIVE.rst  PKG-INFO  README.rst  RELEASING.rst  TIDELIFT.rst  bench  changelog  codecov.yml  doc  extra  pyproject.toml  scripts  setup.cfg  setup.py  src  testing  tox.ini
```
Notice the inclusion of the `testing` directory.

adferrand:
Ok I see with the example of pytest itself. We can have a layout that put the tests out of the python modules, but still include them in the sdist. I used `MANIFEST.in` for that.

The use case describe in certbot `test_sdist.sh` is working properly with this approach.

bmw:
Great! Can we also move src/josepy/testdata out of src/josepy?

adferrand:
For that I may need some help to find an elegant approach to resolve the actual path of the resource. Indeed once `testdata` is moved out of `src`, it is not anymore in the python packages and so `pkg_resources`/`importlib` methods are useless.

So you have some idea on that matter, appart assuming the relative location of `test` folder from the path of the `josepy` package, or using in a way or another the path of the current test module ?

bmw:
Can't we use an approach similar to the one in https://github.com/certbot/certbot/pull/7600?

adferrand:
Indeed, moving test_util and testdata altogether just works. Thanks!

bmw (changes requested):
The only change I'd like to see here is a changelog entry. Maybe something like the following?

> josepy's tests can no longer be imported under the name josepy, however, they are still included in the package and you can run them by installing josepy with "tests" extras and running `python -m pytest`.

bmw (approved):

PR #114, opened 2021-09-14
Allow to add type annotations to fields defined in a `JSONObjectWithFields`

adferrand:
This PR provides a new function, `josepy.field`, that allows to declare fields with appropriate type annotations to express the type of the actual value of the field. An exemple is provided in the associated docstring, and actual usage of it is done in the `jws` module.

### DIFF
--- a/src/josepy/jws.py
+++ b/src/josepy/jws.py
@@ -2,7 +2,7 @@
 import argparse
 import base64
 import sys
-from typing import Dict, Any, Optional, FrozenSet, Mapping, List, Type, cast
+from typing import Dict, Any, Optional, FrozenSet, Mapping, List, Type, Tuple, cast
 
 from OpenSSL import crypto
 
@@ -63,7 +63,7 @@ class Header(json_util.JSONObjectWithFields):
     jwk: jwk_mod.JWK = json_util.field('jwk', decoder=jwk_mod.JWK.from_json, omitempty=True)
     kid: bytes = json_util.field('kid', omitempty=True)
     x5u: bytes = json_util.field('x5u', omitempty=True)
-    x5c: bytes = json_util.field('x5c', omitempty=True, default=())
+    x5c: Tuple[util.ComparableX509, ...] = json_util.field('x5c', omitempty=True, default=())
     x5t: bytes = json_util.field(
         'x5t', decoder=json_util.decode_b64jose, omitempty=True)
     x5tS256: bytes = json_util.field(
@@ -72,7 +72,7 @@ class Header(json_util.JSONObjectWithFields):
                                      decoder=MediaType.decode, omitempty=True)
     cty: MediaType = json_util.field('cty', encoder=MediaType.encode,
                                      decoder=MediaType.decode, omitempty=True)
-    crit: bytes = json_util.field('crit', omitempty=True, default=())
+    crit: Tuple[Any, ...] = json_util.field('crit', omitempty=True, default=())
     _fields: Dict[str, json_util.Field]
 
     def not_omitted(self) -> Dict[str, json_util.Field]:
--- a/src/josepy/jws.py
+++ b/src/josepy/jws.py
@@ -7,7 +7,7 @@ from typing import Dict, Any, Optional, FrozenSet, Mapping, List, Type, cast
 from OpenSSL import crypto
 
 import josepy
-from josepy import b64, errors, json_util, jwa, jwk, util
+from josepy import b64, errors, json_util, jwa, jwk as jwk_mod, util
 
 
 class MediaType:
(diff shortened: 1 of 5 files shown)


bmw:
Did you see https://stackoverflow.com/a/54948342/6534779? An approach like that using generics seems kind of intriguing to me, but there may be downsides with an approach like that which I am not aware of.

adferrand:
Yes I saw these propositions but I found them not convincing enough. Having `type: ignore` is basically ignoring the problem, and playing with metaclass and generics seems quite heavy for a "simple" type problem.

The approach I propose is borrowed from pydantic. On top of my personal appreciation, the fact that it is the approach chosen by a major project claiming to be designed to be type friendly is decisive in my opinion.

bmw:
To be clear, the approach from that post supposedly came from [attrs](https://pypistats.org/packages/attrs) which is much more popular than [pydantic](https://pypistats.org/packages/pydantic). I think the actually relevant code from `attrs` is [here](https://github.com/python-attrs/attrs/blob/f57b6a6759827a0ec56729da27ec39e9e3b006f3/src/attr/__init__.pyi#L123-L221).

My main interest in approach like that is I think the generic may allow `mypy` to automatically determine the type of Fields instead of relying on us to add these type annotations manually. I also noticed that `mypy` doesn't seem to complain if we choose not to annotate a field type with this approach.

What do you think? Does this change your opinion at all? I'm somewhat tempted to have us play with an approach using generics to see what it'd look like before going with the approach here.

adferrand:
Indeed I missed the reference to `attrs`which make it a very valid alternative given the impact of this project.

However I notice that the automated type using helpers and generics was for legacy fields, and that modern fields on Python 3.6+ are using an approach were the type annotation is explicitly defined on the field declaration, which is essentially the same approach than in `pydantic` and what I propose here.

So for whatever reason the `attrs` maintainers dropped the old approach, and so we can suspect it has some drawbacks.

The last proposition in the post is quite interesting, but it goes beyond the purpose of this PR, since it allows to infer the field constructor from the type annotation provided by the user.

Globally I still prefer my approach since it has the minimal impact to how `josepy`can be used, given that the typed field is an opt-in.

bmw:
Forget the stack overflow post for a minute. What do you think about the approach [currently in attrs](https://github.com/python-attrs/attrs/blob/f57b6a6759827a0ec56729da27ec39e9e3b006f3/src/attr/__init__.pyi#L123-L221) which I linked to from my last post?

I played with that a bit locally and there's some things to fix up (such as `mypy` getting mad that `from_json` doesn't return the expected type in inherited objects), but it still seems kinda promising to me personally for the reasons I described in my last post.

adferrand:
After a discussion with @bmw, we agree on adding the minimal modifications on the current design of this PR by enforcing that all fields declared with the new `josepy.field` function are properly annotated with types. If not a `ValueError` will be raised upon declaration of a new `JSONObjectWithFields` subclass.

src/josepy/jws.py
```diff
@@ -57,22 +57,22 @@ class Header(json_util.JSONObjectWithFields):
...
-    jwk = json_util.Field('jwk', decoder=jwk.JWK.from_json, omitempty=True)
-    kid = json_util.Field('kid', omitempty=True)
-    x5u = json_util.Field('x5u', omitempty=True)
-    x5c = json_util.Field('x5c', omitempty=True, default=())
-    x5t = json_util.Field(
+    jku: bytes = json_util.field('jku', omitempty=True)
+    jwk: jwk_mod.JWK = json_util.field('jwk', decoder=jwk_mod.JWK.from_json, omitempty=True)
+    kid: bytes = json_util.field('kid', omitempty=True)
+    x5u: bytes = json_util.field('x5u', omitempty=True)
+    x5c: bytes = json_util.field('x5c', omitempty=True, default=())
```
bmw, line 66: Based on `x5c.{encoder,decoder}` below, I think this should be:
```suggestion
    x5c: Tuple[util.ComparableX509, ...] = json_util.field('x5c', omitempty=True, default=())
```
If you make this change, you'll also need to import `Tuple` from `typing`.

src/josepy/jws.py
```diff
@@ -57,22 +57,22 @@ class Header(json_util.JSONObjectWithFields):
...
-    typ = json_util.Field('typ', encoder=MediaType.encode,
-                          decoder=MediaType.decode, omitempty=True)
-    cty = json_util.Field('cty', encoder=MediaType.encode,
-                          decoder=MediaType.decode, omitempty=True)
-    crit = json_util.Field('crit', omitempty=True, default=())
+    typ: MediaType = json_util.field('typ', encoder=MediaType.encode,
+                                     decoder=MediaType.decode, omitempty=True)
+    cty: MediaType = json_util.field('cty', encoder=MediaType.encode,
+                                     decoder=MediaType.decode, omitempty=True)
+    crit: bytes = json_util.field('crit', omitempty=True, default=())
```
bmw, line 75: With the default value here and the value for `crit.decoder` below, maybe this should be:
```suggestion
    crit: Tuple[...] = json_util.field('crit', omitempty=True, default=())
```
If you make this change, you'll also need to import `Tuple` from `typing`.

bmw (changes requested):
Other than my minor inline comments, can we add a changelog entry about this? Maybe something like the following in the added section:

> Added a field function to assist in adding type annotations for Fields in classes. If the field function is used to define a Field in a JSONObjectWithFields based class without a type annotation, an error will be raised.

src/josepy/jws.py
adferrand: Fixed that.

src/josepy/jws.py
adferrand: Fixed that.

adferrand:
I also take the occasion of this PR to add a changelog entry about https://github.com/certbot/josepy/pull/115.

bmw (approved):
LGTM!

PR #119, opened 2021-10-27
Add generic methods

adferrand:
As a follow-up of https://github.com/certbot/certbot/pull/9073, this PR redefine some methods as generic ones and fix some types.

### DIFF
--- a/src/josepy/json_util.py
+++ b/src/josepy/json_util.py
@@ -219,7 +219,7 @@ class JSONObjectWithFieldsMeta(abc.ABCMeta):
         return abc.ABCMeta.__new__(mcs, name, bases, namespace)  # type: ignore[call-overload]
 
 
-T = TypeVar('T', bound='JSONObjectWithFields')
+GenericJSONObjectWithFields = TypeVar('GenericJSONObjectWithFields', bound='JSONObjectWithFields')
 
 
 class JSONObjectWithFields(util.ImmutableMap,
@@ -331,7 +331,7 @@ class JSONObjectWithFields(util.ImmutableMap,
         return fields
 
     @classmethod
-    def from_json(cls: Type[T], jobj: Mapping[str, Any]) -> T:
+    def from_json(cls: Type[GenericJSONObjectWithFields], jobj: Mapping[str, Any]) -> GenericJSONObjectWithFields:
         return cls(**cls.fields_from_json(jobj))
 
 
--- a/src/josepy/json_util.py
+++ b/src/josepy/json_util.py
@@ -9,7 +9,7 @@ The framework presented here is somewhat based on `Go's "json" package`_
 import abc
 import binascii
 import logging
-from typing import Dict, Type, Any, Callable, List, Mapping, Optional
+from typing import Dict, Type, Any, Callable, List, Mapping, Optional, TypeVar
 
 from OpenSSL import crypto
 import josepy.util
@@ -219,6 +219,9 @@ class JSONObjectWithFieldsMeta(abc.ABCMeta):
         return abc.ABCMeta.__new__(mcs, name, bases, namespace)  # type: ignore[call-overload]
 
 
+T = TypeVar('T', bound='JSONObjectWithFields')
+
+
 class JSONObjectWithFields(util.ImmutableMap,
                            interfaces.JSONDeSerializable,
                            metaclass=JSONObjectWithFieldsMeta):
@@ -328,7 +331,7 @@ class JSONObjectWithFields(util.ImmutableMap,
         return fields
 
     @classmethod
-    def from_json(cls, jobj: Dict[str, Any]) -> 'JSONObjectWithFields':
+    def from_json(cls: Type[T], jobj: Mapping[str, Any]) -> T:
         return cls(**cls.fields_from_json(jobj))
 
 
--- a/src/josepy/util.py
+++ b/src/josepy/util.py
@@ -1,7 +1,7 @@
 """JOSE utilities."""
 import abc
 from types import ModuleType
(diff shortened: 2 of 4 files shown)


src/josepy/json_util.py
```diff
@@ -219,6 +219,9 @@ def __new__(mcs, name: str, bases: List[str],
         return abc.ABCMeta.__new__(mcs, name, bases, namespace)  # type: ignore[call-overload]
 
 
+T = TypeVar('T', bound='JSONObjectWithFields')
```
bmw, line 222: I think/hope no one would be confused by this, but what do you think about prefixing this variable with an underscore to make it clear it's not part of the module's API.

src/josepy/util.py
```diff
@@ -151,6 +151,9 @@ def __hash__(self) -> int:
         raise NotImplementedError()
 
 
+T = TypeVar('T', bound='ImmutableMap')
```
bmw, line 154: I think/hope no one would be confused by this, but what do you think about prefixing this variable with an underscore to make it clear it's not part of the module's API.

bmw (changes requested):
Other than a tiny suggestion about variable names, this LGTM!

After this lands, would you like me to do a release to help with https://github.com/certbot/certbot/pull/9073?

src/josepy/json_util.py
adferrand: In fact I am not sure we should to this. Indeed the object `T` created with `TypeVar` becomes the effective output type of the method trough the annotation declaration. If we declare it as "private", it also means that the output of the public function JSONObjectWithFields.from_json is private, which is not the intended behavior.

I checked in particular that the name of the variable and the `name` parameter passed as first argument in `TypeVar` must be the same, which suppose some significance in the metadata for the runtime. I also fail to find a usage of private `TypeVar` objects in the wild.

At the end I anticipate some linting/mypy errors for external users of `josepy` to have a private element in the signature of the function.

src/josepy/util.py
adferrand: See above.

src/josepy/json_util.py
bmw: > I also fail to find a usage of private TypeVar objects in the wild.

I can find some. I grabbed https://hugovk.github.io/top-pypi-packages/top-pypi-packages-30-days.json and ran this script:
```
#!/usr/bin/env python3
import json
import subprocess

with open('top-pypi-packages-30-days.json') as f:
    j = json.load(f)

for row in j['rows']:
    project = row['project']
    try:
        subprocess.run(f'pip download --only-binary :all: --no-deps {project}'.split())
    except Exception:
        pass
```
If you do this, you may want to modify it to not download all of these packages or just interrupt the program yourself at some point.

I then ran `ls *.whl | xargs -n1 unzip` followed by `grep -Er '\_[^:]+=.+TypeVar' .`. I get 144 matches for this search when looking at the top 100 or so PyPI packages.

> At the end I anticipate some linting/mypy errors for external users of josepy to have a private element in the signature of the function.

At the end for me though, can you point me to a realistic use case you currently have in mind where this variable needs to be public? If so, we can probably make it public. If not, I propose we make it private and if it's ever needed in the future, we can make it public.

adferrand: If I reuse your example, in fact from a similar output (146 entries), when I deduplicate the python/stubs files (py/pyi) and I remove the private TypeVar in private modules, the number is in fact 28 entries. It could drop a little more for situation where the private `TypeVar` is used in private functions or private classes, but I have not went that far in the analysis.

It still feels wrong to me to use a private `TypeVar` for the public API for the following reasons.

1. I am certainly influenced by the way of doing things in Java in particular, but I consider these types as part of the signature of the method. In that extent, using private fields (eg. a private inner static class) in the signature of public method is a bad design in Java, as it makes impossible to override this method since you cannot refer to its private components in the signature.
2. Another way to see things is that if we would relate the type annotations with the docstring, we would need to refer to some private elements in the documentation, or find a workaround to express that with only public components.
3. I relate this discussion to another one done on certbot concerning the display test utils, where it was a problem for you to expose some internal components name in the signature of `patch_display_util()`  function if I remember correctly.
4. If we generate the stubs files with `stubgen src` while using a private `TypeVar`, it turns out that the stub file `json_util.pyi` is syntactically wrong because `_T` variable is not generated in the file, while `from_json()`  declares to return `_T`. We will not use the stub files but the implicit type resolution from source code, and I guess mypy is still capable to workaround that. But this shows that the choice of making the `TypeVar` private is not neutral and may create unexpected behaviors.

bmw: OK. What do you think about instead giving these types more descriptive names than `T`, but keeping them public?

adferrand: About about `GenericJson`, `GenericJsonObject`or `GenericJsonObjectWithFields` ?

bmw: I don't care too much, but I kind of like the pattern `Generic<ClassName>` so in this case it'd be `GenericJSONObjectWithFields` and `GenericImmutableMap`.

adferrand: Ok, done with `GenericJSONObjectWithFields`.

src/josepy/util.py
adferrand: Done with `GenericImmutableMap`.

adferrand:
To answer your general comment @bmw, yes if you can create a new version of Josepy right after this PR this would be great. I will then update the related PR on Certbot to avoid the current workarounds, and update with the same naming strategy the `TypeVar` instances added there.

bmw (approved):

bmw:
Sounds good. Josepy 1.11.0 has been released.

PR #120, opened 2021-11-09
Add rtd reqs

bmw:
This is the josepy version of https://github.com/certbot/certbot/pull/9083.

You can see the latest build failing [here](https://readthedocs.org/projects/josepy/builds/15187466/) and the build passing on this branch [here](https://readthedocs.org/projects/josepy/builds/15219421/).


alexzorin (commented):
1. >  Examine the produced list and update invididual entries here.

     What's the reason to do this, rather than just dumping `pip freeze` in here and looking at the diff?
2. Do you think it would help to keep this list sorted, so it's a bit easier to see what's changed?

bmw:
1. For reasons like those described in https://github.com/certbot/josepy/issues/101.

2. The file in this PR was previously sorted using the command line utility `sort`. I did this because I updated it by combining the `pip freeze` output from multiple systems. This sorted things slightly differently though so I resorted things the way they were before by installing all of these packages in a single virtual environment and running `pip freeze`.

alexzorin (approved):
Thanks for the explanation.

PR #122, opened 2021-11-18
Make several methods of TypedJSONObjectWithFields generic

adferrand:
See https://github.com/certbot/certbot/pull/9073#discussion_r752156133


adferrand:
@bmw I have second thoughts about the naming pattern `Generic<ClassName>` for the public `TypeVar` instances introduced in the project. Given that the `TypeVar` itself is not a generic, but the variable of a generic representing "any" possible type of the given bound, I think a better name would be `Any<ClassName>` (eg. `AnyJSONObjectWithField`, `AnyImmutableMap`).

But if is too late, I can live with `Generic<ClassName>` and prefer to keep a coherent convention for `GenericTypedJSONObjectWithFields`.

bmw:
To respond to https://github.com/certbot/certbot/pull/9073#discussion_r752156133:

I'm torn. Looking at our code, I agree with you that's how we use it, but there's no reason it has to be used that way does it? Does the class returned by `get_type_cls` have to be a subclass of the current class or couldn't it be an unrelated class?

adferrand:
Well not totally any object. It must be something that have a `fields_from_json` static method accepting a unique parameter of type `Dict[str, Any]` and whose return type is a `Mapping[str, Any]` whose destructuration is an acceptable input of its own `__init__` method too create an instance.

We could define a protocol for that.

However I think that the typing in python van express the intention of usage and not necessarily the possible usages since it is still possible to ignore the warnings on caller side if "it works".

In that extent I think we can claim to accept a  subclass of `TypedJSONObjectWithFields`, or maybe `JSONObjectWithFields`.

adferrand:
In the light of https://github.com/certbot/certbot/pull/9073#discussion_r755468894, I mark also all fields using `omitempty=true` without default value as `Optional[...]` in `jose.Header`.

adferrand:
In the light of https://github.com/certbot/certbot/pull/9073#discussion_r755470783, I redeclare `kid` in `jose.Header` as `Optional[str]`.

bmw:
> Well not totally any object. It must be something that have a `fields_from_json` static method accepting a unique parameter of type `Dict[str, Any]` and whose return type is a `Mapping[str, Any]` whose destructuration is an acceptable input of its own `__init__` method too create an instance.
>
> We could define a protocol for that.
>
> However I think that the typing in python van express the intention of usage and not necessarily the possible usages since it is still possible to ignore the warnings on caller side if "it works".
>
> In that extent I think we can claim to accept a  subclass of `TypedJSONObjectWithFields`, or maybe `JSONObjectWithFields`.

Thanks for clarifying. There's nothing that requires the return value of `from_json` to inherit from the specific `TypedJSONObjectWithFields` (sub)class in use though is there? For instance, in `src/josepy/jwk.py` we declare a `JWK` class and register a number of its subclasses with it. This is how we always use `TypedJSONObjectWithFields`, but there's nothing that requires or even states in something like documentation that the registered class should be a subclass of `JWK` is there?

If not, I personally think the types we can declare here without enforcing our specific and somewhat undocumented usage of them is pretty limited and won't really help with https://github.com/certbot/certbot/pull/9073#discussion_r752156133. If that's the case, I personally think we should just drop this piece of the PR and continue to use `cast` there. I think `cast` is a code smell, but I also think sometimes it's the best/simplest option.

What do you think?

adferrand:
Yes basically this was my first intention when trying to address the `cast` required on acme side, which was basically, we cannot do anything about it because the function here may return any json class here.

After that I saw that the effective usage of `josepy` could authorize an explicit limitation of what we claim to be the intended usage of this method thanks to generics and remove the cast.

As I said previously on a statically typed language, the types are quite naturally about what a function can do and can accept. On the situation of mypy with python I thought that we could have a slight room to say something a little different: what is the intended usage, since at the end, the runtime is dynamic and objects just need to comply with structural typing (duct typing).

Now if your point here, and it is more about the usual definition of a type.

So if you agree, I will keep a type accepting or returning any subclass of `TypedJSONObjectWithFields` without relation to the actual type of the object on which the methods are called.

On a more general observation here, I think the design of a dedicated and independant factory function at the module level would have made more sense here, since attaching a function to the class do not add a specific value at the end.

constraints.txt
```diff
@@ -7,58 +7,73 @@
...
+coverage==5.5
+cryptography==3.4.8
+dataclasses-json==0.5.5
+distlib==0.3.2
+docutils==0.17
+filelock==3.0.12
+flake8==3.9.2
+idna==3.2
 imagesize==1.2.0
-importlib-metadata==4.2.0
```
bmw, line 24: I think pinnings like this are still needed. See https://github.com/certbot/josepy/pull/122/checks#step:5:152.

What I personally have done when updating this file is to combine the output of `pip freeze` from the newest and oldest versions of Python we support. See https://github.com/certbot/josepy/issues/101 for more about this including considering switching to `poetry` here.

src/josepy/json_util.py
```diff
@@ -457,6 +457,9 @@ def decode_csr(b64der: str) -> josepy.util.ComparableX509:
         raise errors.DeserializationError(error)
 
 
+GenericTypedJSONObjectWithFields = TypeVar('GenericTypedJSONObjectWithFields', bound='TypedJSONObjectWithFields')
```
bmw, line 460: With it's new restricted usage, does this generic still provide value for us? If so, can you give me an example of what that is?

There very well may be a good reason to keep it, but since it's only used on non-self and non-class parameters and on return values, it's not immediately clear to me what its value is anymore.

constraints.txt
```diff
@@ -7,58 +7,73 @@
...
-tox==3.24.4
-typed-ast==1.4.3
+tox==3.24.3
+types-cryptography==3.3.5
+types-enum34==1.1.0
+types-ipaddress==1.0.0
+types-pyOpenSSL==20.0.6
+types-pyRFC3339==0.1.2
+types-requests==2.25.6
+types-setuptools==57.0.2
```
bmw, line 75: This works in CI because we currently run `pip install -r constraints.txt`, but `tox -e mypy` fails for me locally.

I think we should fix this by declaring a dependency on these `types-*` packages in `setup.py` like we do in Certbot.

src/josepy/jwk.py
```diff
@@ -85,7 +85,7 @@ def _load_cryptography_key(cls, data: bytes, password: Optional[bytes] = None,
         for loader_public in (serialization.load_pem_public_key,
                               serialization.load_der_public_key):
             try:
-                return loader_public(data, backend)
+                return loader_public(data, backend)  # type: ignore[operator]
```
bmw, line 88: Is it possible/desired to declare a type for `loader_public` to avoid this `ignore`?

src/josepy/jwa.py
```diff
@@ -181,7 +181,7 @@ def _sign(self, key: ec.EllipticCurvePrivateKey, msg: bytes) -> bytes:
         try:
             if new_api:
                 return key.sign(msg, ec.ECDSA(self.hash))
-            signer = key.signer(ec.ECDSA(self.hash))
+            signer = key.signer(ec.ECDSA(self.hash))  # type: ignore[attr-defined]
```
bmw, line 184: Do you know what's going on with the `ignore` lines here? It seems like these attributes exist. What should ideally be "fixed" so these ignores are no longer required in your opinion?

src/josepy/jws.py
```diff
@@ -57,21 +57,21 @@ class Header(json_util.JSONObjectWithFields):
     :ivar str cty: Content-Type, inc. :const:`MediaType.PREFIX`.
 
     """
-    alg: jwa.JWASignature = json_util.field(
+    alg: Optional[jwa.JWASignature] = json_util.field(
         'alg', decoder=jwa.JWASignature.from_json, omitempty=True)
-    jku: bytes = json_util.field('jku', omitempty=True)
+    jku: Optional[bytes] = json_util.field('jku', omitempty=True)
     jwk: jwk_mod.JWK = json_util.field('jwk', decoder=jwk_mod.JWK.from_json, omitempty=True)
```
bmw, line 63: Since no `default` is given here, should this type also be `Optional`?

src/josepy/jws.py
```diff
@@ -57,21 +57,21 @@ class Header(json_util.JSONObjectWithFields):
...
-    kid: bytes = json_util.field('kid', omitempty=True)
-    x5u: bytes = json_util.field('x5u', omitempty=True)
+    kid: Optional[str] = json_util.field('kid', omitempty=True)
+    x5u: Optional[bytes] = json_util.field('x5u', omitempty=True)
     x5c: Tuple[util.ComparableX509, ...] = json_util.field('x5c', omitempty=True, default=())
-    x5t: bytes = json_util.field(
+    x5t: Optional[bytes] = json_util.field(
         'x5t', decoder=json_util.decode_b64jose, omitempty=True)
     x5tS256: bytes = json_util.field(
```
bmw, line 69: Since no `default` is given here, should this type also be `Optional`?

bmw (changes requested):

constraints.txt
adferrand: Do not ask me twice for poetry ;)

In the mean time, I upgraded the contraint file using your method (Python 3.6 and Python 3.10).

src/josepy/json_util.py
adferrand: Yes, it is still useful for the `register()` function. It helps mypy to understand that a subclass of `TypedJSONObjectWithFields` annotated with the decorator `@register` is still this subclass, and not `TypedJSONObjectWithFields` itself.

constraints.txt
adferrand: Added.

src/josepy/jwk.py
adferrand: Well, with this design it is difficult to do better. I would like to add a type hint to say it is a `Callable`, but mypy still complains it is a `function`. I changed to a slightly better version that it is to say `loader_private: Any`, which is still ignoring the error, but with a proper annotation instead of disabling the error.

src/josepy/jwa.py
adferrand: Yes basically this code relates to old versions of cryptography where `sign` function was not existing. `signer` and `verifier` methods have been deprecated since cryptography 2.1.

It seems that the `types-cryptography` stub package has been proactive and these functions have already been removed, certainly related to the final deprecation warning made by cryptography that 37.0.0 will remove these functions, in order to push developers to stop using them.

The simplest fix I think it to declare dependency on `cryptography>=1.5` instead of `cryptography>=0.8`, and stop using this retro-compatible branch.

src/josepy/jws.py
adferrand: Totally! I fixed that.

src/josepy/jws.py
adferrand: Totally! I fixed that.

bmw (changes requested):
> @bmw I have second thoughts about the naming pattern `Generic<ClassName>` for the public `TypeVar` instances introduced in the project. Given that the `TypeVar` itself is not a generic, but the variable of a generic representing "any" possible type of a given bound, I think a better name would be `Any<ClassName>` (eg. `AnyJSONObjectWithField`, `AnyImmutableMap`).
>
> But if is too late, I can live with `Generic<ClassName>` and prefer to keep a coherent convention for `GenericTypedJSONObjectWithFields`.

Sorry I missed this question earlier. I personally don't love the `Any*` name. In this PR, the return value for `register` isn't **any** `TypedJSONObjectWithFields` and is instead the specific subclass passed in as the parameter. Maybe `TypedJSONObjectWithFieldsGeneric`, `TypedJSONObjectWithFieldsType`, or `TypedJSONObjectWithFieldsTypeVar` is better?

I personally am in favor of just sticking with our current convention though. It's not perfect, but I think the name "Generic" helps communicate it is a Python typing generic and including the class name helps communicate the bounds of the type.

src/josepy/jwa.py
adferrand: Works for me ! Cryptography 1.5 has been added 5 years ago, I think it is fine in term of backward compatibility.

adferrand: Dependencies and constraints updated.

constraints.txt
```diff
@@ -75,5 +66,9 @@ types-requests==2.25.6
 types-setuptools==57.0.2
 typing-extensions==3.10.0.2
 typing-inspect==0.7.1
-urllib3==1.26.6
-virtualenv==20.7.2
+typed-ast==1.4.3
+# typing_extensions should be pinned, but this will conflicts with typing-extensions==3.10.0.2
```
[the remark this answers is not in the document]
adferrand: I do not know what happened, but the inconsistency of the pinning vanished with the latest update after bumping cryptography.

adferrand:
I agree for to keep the current naming of the `TypeVars`.

src/josepy/jwa.py
```diff
@@ -105,36 +105,20 @@ class _JWARSA:
...
-        signer.update(msg)
-        try:
-            return signer.finalize()
-        except ValueError as error:
-            logger.debug(error, exc_info=True)
-            raise errors.Error(str(error))
 
     def verify(self, key: rsa.RSAPublicKey, msg: bytes, sig: bytes) -> bool:
         """Verify the ``msg` and ``sig`` using ``key``."""
         # If cryptography library supports new style api (v1.4 and later)
```
bmw, line 119:
```suggestion
```

src/josepy/jwa.py
```diff
@@ -105,36 +105,20 @@ class _JWARSA:
     def sign(self, key: rsa.RSAPrivateKey, msg: bytes) -> bytes:
         """Sign the ``msg`` using ``key``."""
         # If cryptography library supports new style api (v1.4 and later)
```
bmw, line 107:
```suggestion
```

src/josepy/jwa.py
```diff
@@ -209,15 +183,8 @@ def verify(self, key: ec.EllipticCurvePublicKey, msg: bytes, sig: bytes) -> bool
 
     def _verify(self, key: ec.EllipticCurvePublicKey, msg: bytes, asn1sig: bytes) -> bool:
         # If cryptography library supports new style api (v1.4 and later)
```
bmw, line 185:
```suggestion
```

bmw (changes requested):
I found some outdated comments but this otherwise LGTM!

bmw (approved):

PR #124, opened 2021-11-30
Test all branches

bmw:
Prior to https://github.com/certbot/josepy/pull/104, CI was configured to run tests on pushes to all branches. I personally find this setup useful as I've since resorted to things like [opening a draft PR if I want to test things in CI](https://github.com/certbot/josepy/pull/123).

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -49,8 +49,9 @@ jobs:
         with:
           fail_ci_if_error: true
   notify:
-    # Only notify about failed builds for branches (not PRs).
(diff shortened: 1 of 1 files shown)


alexzorin (commented):
Looks like this will notify MM on non-master branch failures? Is that intentional?

bmw:
Nice catch!

I fixed this by reusing the config from https://github.com/certbot/website/blob/e4db635ed89081fae5c8ce1bdf6ebf7bd7e6e0e6/.github/workflows/ci.yml#L59-L61.

alexzorin (approved):

PR #127, opened 2022-01-29
Fix from_json and json_loads typehints

mathiasertl:
Fix two type hinting issues that I stumbled upon:

* Use generics for `from_json()` and `json_loads()`. This way the return value in a subclass is (correctly) typehinted to that subclass. See also: https://mypy.readthedocs.io/en/stable/generics.html#generic-methods-and-generic-self
* `json.loads()` also accepts bytes, so `json_loads()` actually does too.

### DIFF
--- a/src/josepy/interfaces.py
+++ b/src/josepy/interfaces.py
@@ -10,7 +10,7 @@ from collections.abc import Sequence, Mapping
 # pylint: disable=no-self-argument,no-method-argument,no-init,inherit-non-class
 # pylint: disable=too-few-public-methods
 
-JSONDeSerializableTypeVar = TypeVar("JSONDeSerializableTypeVar", bound="JSONDeSerializable")
+GenericJSONDeSerializable = TypeVar("GenericJSONDeSerializable", bound="JSONDeSerializable")
 
 
 class JSONDeSerializable(object, metaclass=abc.ABCMeta):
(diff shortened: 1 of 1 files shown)


mathiasertl:
I am  a bit confused about linting in this project. Flake8 does not seem to complain about any line length for some reason, but I could not find any configuration for that either. If there's some linting requirements, please let me know, I'll of course adapt.

src/josepy/interfaces.py
```diff
@@ -10,6 +10,8 @@
 # pylint: disable=no-self-argument,no-method-argument,no-init,inherit-non-class
 # pylint: disable=too-few-public-methods
 
+JSONDeSerializableTypeVar = TypeVar("JSONDeSerializableTypeVar", bound="JSONDeSerializable")
```
adferrand, line 13: To be compliant with the other generics introduced in the project, could you rename this TypeVar into `GenericJSONDeSerializable` ?

adferrand (changes requested):

bmw:
Thanks Adrien.

I'm unassigning myself and unfollowing this PR for now, but by all means ping me if I should take a look.

src/josepy/interfaces.py
mathiasertl: Certainly! done in last commit.

adferrand (approved):
Great, thanks a lot !

PR #128, opened 2022-02-04
Introduce linting (flake8 and isort)

atombrella:
I think it'd make sense to activate `flake8` as well as introducing a standardized way of importing things.

There are still various comments relating to how pylint should work, but it doesn't seem to be used. Maybe those could be deleted/evaluated at some point. 

`E501` (maximum line length) should also be fixed. There aren't many places where it's wrong.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -41,6 +41,11 @@ jobs:
       # STATIC CHECK
       - name: Run mypy
         run: tox -e mypy
+      # LINTING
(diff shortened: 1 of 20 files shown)


adferrand (approved):
All of what you are presenting here seems very decent. I was initially hesitant to enforce a specific format for the imports, but in the scope of the potential introduction of black in our projects, and given the good practices I could see in other projects regarding this topic (in particular the go-based projects), let's give it a try!

PR #129, opened 2022-02-06
Migrate the project to Poetry

adferrand:
This PR is a proposition to migrate the project josepy to poetry, the alternative to setuptools/pip already used in Certbot to calculate the pinning dependencies of the project. Given that josepy is a a mono-project in one repository, it is a good candidate to take full advantage of poetry without major disruption.

To do this migration, a `pyproject.toml` is created that merges configuration files of tools that currently support this format: `.coveragerc`, `mypy.ini`, most of `setup.cfg` and of course `setup.py`.

The configuration concerning setuptools is mostly taken as it was on `setup.py`, with two noticeable difference: the extras `dev` and `tests` are merged into the built-in dev dependencies management in Poetry, and so are not exposed anymore as extras to the published artifacts in PyPI. Also, flak8 is not run anymore as part of pytest (since there is a dedicated tox target, and this removes several workaround in the dependencies and warning ignores.

The file `CONTRIBUTING.md` is updated to add install and setup sections for developers.

Tox configuration is updating accordingly.

The `tools/release.sh` script is adapted. I encountered some difficulties however to get it right. In particular I have not understood the actual purpose of the tests done at the end of the script, which seems to not do anything actually: no test is run with `pytest --pyargs josepy` (except the flake8 checks that are removed by this PR).


setup.cfg
```diff
@@ -1,15 +1,4 @@
-[bdist_wheel]
-universal = 1
-
 [flake8]
 exclude = .git,.tox
 ignore = W504
 max-line-length = 100
```
atombrella, line 4: Mainly just a question for debate, but could this be moved to `.flake8` to signal that everything has been moved to Poetry?

atombrella (commented):
Not at all familiar with poetry, but I have a few questions.

bmw (changes requested):
This isn't quite a full review, but this it's close to one. Thanks for picking this up Adrien and giving me an excuse to play more with poetry.

The remaining things I want to look at are:

1. How do the packages built by the release script in this PR differ from the ones in `master`? I've ran the script both ways locally, extracted the tarball & wheel contents, but I haven't gone through the diff yet.
2. On that note, will this break anything for packagers or users? Similar to our "oldest" tests on the Certbot repo, will EPEL 8 have any trouble packaging this?
3. We should handle [this](https://github.com/certbot/josepy/blob/15e9c508252381e636e0159a0795bc3f22fa952d/constraints.txt#L6-L7) somehow.

As a final note for myself, I should check if this changes our (undocumented) release process in any significant way.

pyproject.toml
```diff
@@ -0,0 +1,107 @@
...
+branch = true
+source = ["josepy"]
+
+[tool.coverage.paths]
+source = [
+   ".tox/*/lib/python*/site-packages/josepy",
+   ".tox/pypy*/site-packages/josepy",
+]
+
+[tool.coverage.report]
```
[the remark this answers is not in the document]
adferrand: No it is an override of the default behavior (which is to ignore lines commented with `pragma: no cover`).

setup.cfg
adferrand: I have good hope that flake will add its support to pyproject.toml, so I think we can keep things like this until the support is added.

pyproject.toml
```diff
@@ -0,0 +1,107 @@
...
+# add sign() and verify() to asymetric keys (RSA >=1.4, ECDSA >=1.5)
+cryptography = ">=1.5"
+# Connection.set_tlsext_host_name (>=0.13)
+pyopenssl = ">=0.13"
+# For pkg_resources. >=1.0 so pip resolves it to a version cryptography
+# will tolerate; see #2599:
+setuptools = ">=1.0"
+
+[tool.poetry.dev-dependencies]
+coverage = {version = ">=4.0", extras = ["toml"]}
```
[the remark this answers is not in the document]
adferrand: It makes coverage capable to read its configuration from `pyproject.toml`. I added a comment about that.

pyproject.toml
```diff
@@ -0,0 +1,107 @@
...
+types-pyOpenSSL = "*"
+types-pyRFC3339 = "*"
+types-requests = "*"
+types-setuptools = "*"
+pytest = ">=2.8.0"
+pytest-cov = "*"
+pytest-flake8 = ">=0.5"
+tox = "*"
+sphinx = {version = ">=1.0", optional = true}  # autodoc_member_order = 'bysource', autodoc_default_flags
+sphinx-rtd-theme = {version = ">=1.0", optional = true}
```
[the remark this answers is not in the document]
adferrand: You are right. Setting optional dependencies for extra in dev is undocumented.

pyproject.toml
```diff
@@ -0,0 +1,107 @@
...
+flake8 = "*"
+isort = "*"
+mypy = "*"
+types-pyOpenSSL = "*"
+types-pyRFC3339 = "*"
+types-requests = "*"
+types-setuptools = "*"
+pytest = ">=2.8.0"
+pytest-cov = "*"
+pytest-flake8 = ">=0.5"
```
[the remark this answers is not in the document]
adferrand: Indeed!

tox.ini
```diff
@@ -1,22 +1,25 @@
...
 [testenv]
+whitelist_externals = poetry
 commands =
-    py.test {posargs}
-deps =
-    -cconstraints.txt
-    -e .[tests]
+    poetry install -v
+    poetry run pytest {posargs}
+    poetry run coverage xml
```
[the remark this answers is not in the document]
adferrand: I took it from the code in master, but indeed it seems redundant. Let's remove the command and see.

.github/workflows/check.yaml
```diff
@@ -27,16 +27,16 @@ jobs:
...
+          # Look to see if there is a cache hit for the corresponding lock file
+          key: ${{ runner.os }}-poetry-${{ hashFiles('poetry.lock') }}
           restore-keys: |
-            ${{ runner.os }}-pip-
+            ${{ runner.os }}-poetry-
             ${{ runner.os }}-
-      - name: Install Dependencies
-        run: pip install -r constraints.txt
+      - name: Install Poetry & Tox
+        run: pip install poetry>1.0.0 tox>3.3.0
```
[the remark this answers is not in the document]
adferrand: My opinion on this is, out of a locking mechanism, we should follow the usual pattern we have in dependencies constraints, and define only the lower bound and not the upper bound if we do not have a specific reason to cap the dependencies.

This way we transparently benefit from the improvement made by the community, if we can suppose that modern versions breaking our pipeline will be extremely rare.

I never saw tox breaking anything. Poetry is younger so it is difficult to assess.

So far I would keep things like in this PR reconsider that the breaking changes happen regularly.

pyproject.toml
```diff
@@ -0,0 +1,107 @@
...
+
+# Poetry tooling configuration
+
+[tool.poetry]
+name = "josepy"
+version = "1.12.0.dev0"
+description = "JOSE protocol implementation in Python"
+license = "Apache License 2.0"
+classifiers = [
+    "Development Status :: 3 - Alpha",
```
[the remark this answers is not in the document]
adferrand: I think we can take the opportunity to claim that indeed.

cc @bmw

adferrand:
Here are some feedbacks I can share regarding the usage of poetry in a public repo:

For users (here, the developers using the library in their own project), I expect no disturbance. In Lexicon, when I switched to a build of the artifacts by poetry, it seems that nobody noticed a difference since I received no issue about it. For all that matters in that purpose, the tarballs and wheels that poetry generates are standard-compliant with the couple distutils/setuptools.

For maintainers, it depends. I received remarks from some of them when Lexicon has been switched to poetry. The problem was not poetry itself, but the fact that the project build logic becomes driven by PEP-517 and PEP-518. At the end the integrators updated their tooling to support the isolated builds and everything worked. So far I have not received complaints from DEB and RPM environments. It seems that for Lexicon, RPM tools are compliant with PEP-517 and are used for Lexicon (https://src.fedoraproject.org/rpms/python-dns-lexicon/blob/rawhide/f/python-dns-lexicon.spec, https://src.fedoraproject.org/rpms/pyproject-rpm-macros). I expect that now, all main integration tools are compliant with isolated builds.

bmw (commented):
I think this is very close!

One problem I found is I think we still need to migrate and delete `MANIFEST.in`. In particular, I think the changelog, docs, and tests should be at least included in the `sdist` (see https://python-poetry.org/docs/pyproject/#packages). This is to allow OS packagers to easily see the changes, [build the docs](https://salsa.debian.org/letsencrypt-team/certbot/josepy/-/blob/master/debian/control#L37), and [run tests](https://src.fedoraproject.org/rpms/python-josepy/blob/rawhide/f/python-josepy.spec#_92).

I'd appreciate your thoughts on how we expect people to build docs and run tests though. For example, I noticed the docs extras are no longer being installed in the dev environment which I think we should fix. Additionally, how do we expect packagers to run tests? What about [cryptography](https://github.com/pyca/cryptography/blob/89c7ba01ca00fafbddcab2c4ea5be90411fc0277/.github/downstream.d/certbot-josepy.sh)?

For cryptography, I'd like to hear how you recommend migrating them to the new system. I think ideally we'd create one PR that would land before this is merged to stop things from breaking and then create another PR that'd use the `poetry.lock` file.

Depending on the answers to these questions, I think we may also want to include a changelog entry but this otherwise LGTM!

adferrand:
I fixed the inclusion of these files in the sdist.

For your question about tests and docs, I think it is a little complex and I would like to propose a grid to analyze this first.

Let's consider that we have two kind of developers acting on the project: the insiders, who are the contributors working directly on the project, and the outsiders, integrators or developers using the project for their own projects.

By heart, the development tool used in the project (here Poetry introduced by this PR) is primarly designed for insiders: it gives a venv management and dev dependencies that are designed exclusively to develop the project. PEP-517 (that Poetry supports), by providing an isolated build environment, is typically for outsiders since it helps having a tooling system as close as possible to the insiders.

The question now is: is Poetry also suitable for outsiders also ?

If yes:
* installing an environment with the docs can be done using `poetry install -E docs`
* alternatively we directly include the docs dependencies inside the dev dependencies and drop the extra `docs`
* integrator use directly poetry to build the OS packages

If no:
* outsiders must use tools compatible with PEP-517
* installing dependencies for `docs` must take advantage of the extra mechanism, with something like `pip install .[docs]` without the editable flag, which is not supported by pip for now for PEP-517 compliant projects
* running tests must follow a global convention which is to rely on having pytest available, and nothing specific that would be described in the poetry dev dependencies since it would not be accessible to outsiders that are not using poetry: it seems that this convention is followed by the RPM tooling system for pyproject-based projects: https://src.fedoraproject.org/rpms/pyproject-rpm-macros/blob/rawhide/f/pyproject-rpm-macros.spec

For the widest compatibility purpose, I think we should not force outsiders to use poetry, and I would be more in favor of the second approach.

So in my opinion we need to add the following actions:
* update the `CONTRIBUTING.md` to use `poetry install -E docs` instead of `poetry install` for insiders
* keep the `docs` extra so that outsiders can build the docs using `pip install .[docs]`
* open a PR for cryptography to basically do: `pip install . pytest && pytest tests`

On the last point, these modifications ensure that the environment will work before and after this PR is merged and a new release is done. After that, we can also propose to cryptography maintainers to use directly poetry if they want to take advantage of `poetry.lock`.

What do you think of this analysis and plan?

pyproject.toml
```diff
@@ -0,0 +1,117 @@
...
+    "Topic :: Internet :: WWW/HTTP", 
+    "Topic :: Security",
+]
+homepage = "https://github.com/certbot/josepy"
+authors = ["Certbot Project <certbot-dev@eff.org>"]
+readme = "README.rst"
+include = [
+    "CHANGELOG.rst",
+    "CONTRIBUTING.md",
+    ".coveragerc", "pytest.ini", "tox.ini",
```
bmw, line 34: `.coveragerc` and `pytest.ini` no longer exist and while this is arguably out of scope of this PR, I see no reason for us to continue to include `tox.ini`.
```suggestion
```

bmw (changes requested):
Thanks a lot for doing that analysis Adrien. I like your plan a lot. If we go this route, I'm not sure we really need to add a changelog entry like I suggested since it's basically just an internal change.

Another idea for `cryptography` would be to have them pin an old version of `josepy` in their tests temporarily until this PR lands and we can switch them over to `poetry`. They're friendly and easy to work with though so I imagine we could pretty easily come up with something everyone is happy with.

The only thing I don't love about this plan is that we never test our packages the way our packagers do. By that I mean we always install all of our dev extras and use the pytest settings in `pyproject.toml` and never just run the tests with the minimal dependencies plus `pytest` without `pyproject.toml`. I worry this could allow uncaught problems to slip through.

What do you think? Think it's worth doing something like adding a CI run to test this case?

adferrand:
Indeed having a dedicated test to run them as integrators seems a good idea. I added that to the pipeline.

For cryptography, we can check with them if they are willing to use poetry. But at the minimum, using pip in non-editable mode will do the trick.

adferrand:
Update: with `python-core` 1.0.8 released on one side (https://github.com/python-poetry/poetry-core/releases/tag/1.0.8), and along with `pip` 21.3, PEP-660 (https://peps.python.org/pep-0660/) is fully supported and josepy can be installed in editable mode with pip.

I updated the `build-dependencies` accordingly to required at least `python-core` 1.0.8.

If cryptography wants to take advantage of the locked file as they were using `constraints.txt` before, however they will need to use poetry indeed.

.github/workflows/check.yaml
```diff
@@ -51,11 +51,23 @@ jobs:
         uses: codecov/codecov-action@v2
         with:
           fail_ci_if_error: true
+  external:
```
bmw, line 54:
```suggestion
    # This job runs our tests like external parties such as packagers.
    external:
```

tools/release.sh
```diff
@@ -125,6 +126,7 @@ pip install -U pip
 pip install \
   --no-cache-dir \
   --extra-index-url http://localhost:$PORT \
+  --constraint constraints.txt \
```
bmw, line 129: When I hit this line locally the script crashes with:
```
ERROR: Invalid requirement: 'Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.' (from line 1 of constraints.txt)
```
Also, can we move `constraints.txt` out of the `dist.$version` directory? That currently only contains things we upload to PyPI and I think it'd be nice to keep it that way.

bmw (changes requested):
Other than my inline suggestions, we should get a PR against cryptography merged before we land this. Are you up for writing it or should I?

bmw:
Also, sorry for the delayed rereview. I was on vacation but I'm back now.

tools/release.sh
adferrand: Indeed the lock file was outdated. This is fixed. I also moved the generated constraints.txt file to upper level. I also added it to `.gitignore` so that it is never committed, but I can also modify the `release.sh` script to put the file in a disposable temporary directory if you prefer.

adferrand:
For the strategy on cryptography side. So the test indicates they are interested to install josepy with the exact constraints provided in the upstream repository. Which would imply to make them move explicitly to poetry to achieve the same purpose.

In this case we would go more on your approach, which is to do a first PR before this one is merged to pin their version to the current one, then a second one when josepy is release to move their test to poetry.

What do you think of this plan?

tools/release.sh
```diff
@@ -136,7 +127,7 @@ pip install \
   --no-cache-dir \
   --extra-index-url http://localhost:$PORT \
   --constraint ../constraints.txt \
-  josepy[tests]
+  josepy
```
bmw, line 130: Now when I run this script, this line crashes with:
```
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    josepy from http://localhost:1234/josepy/josepy-99.0.0-py3-none-any.whl
```
It'd be nice to somehow add josepy's hash here, but barring you having some elegant idea on how to do this, I think we should run `poetry export` with `--without-hashes`.

tools/release.sh
```diff
@@ -151,6 +142,7 @@ fi
 mkdir kgs
 kgs="kgs/$version"
 pip freeze | tee $kgs
+pip install pytest
```
bmw, line 145: I noticed that with this change `pytest` is no longer pinned. To fix that, how about we include `--dev` in the `poetry export` command and reuse the constraints file here. What do you think?

tools/release.sh
```diff
@@ -117,14 +107,15 @@ done
 
 mkdir "dist.$version"
 mv dist "dist.$version/josepy"
+poetry export -f requirements.txt > constraints.txt
```
bmw, line 110: I think adding this to `.gitignore` like you suggested [here](https://github.com/certbot/josepy/pull/129#discussion_r845571073) is just fine, but it looks like that change has it made its way to this branch yet.

constraints.txt
```diff
@@ -1,73 +0,0 @@
-# Specifies Python package versions for testing.
```
bmw, line 1: One thing I just happened to notice is this file is referenced in `docs/requirements.txt` which is used by readthedocs. We should fix this somehow, but I'm not immediately sure the best way to do it. I think we should keep our dependencies pinned on readthedocs if at all possible.

bmw (changes requested):
> For the strategy on cryptography side. So the test indicates they are interested to install josepy with the exact constraints provided in the upstream repository. Which would imply to make them move explicitly to poetry to achieve the same purpose.
> 
> In this case we would go more on your approach, which is to do a first PR before this one is merged to pin their version to the current one, then a second one when josepy is release to move their test to poetry.
> 
> What do you think of this plan?

This sounds good to me! Are you up for writing these PRs or would you like me to do it?

tools/release.sh
adferrand: Indeed the commit was pending on my local repo. It is fixed.

tools/release.sh
adferrand: Totally, I do that!

tools/release.sh
adferrand: Well, in theory we can just calculate the hash ourselves from the assets generated by `poetry build`, then expose these hashes in the local pypi server. But it sounds like a lot of complexity for not that much of gains. Very unfortunate that pip do not support (yet?) a mix of hashes/unhashes dependencies.

I guess the best for now is to remove the other hashes. I do that.

constraints.txt
adferrand: This was surprisingly easy. Thanks to the `.readthedocs.yaml` file that you can put at the root of the repo, you can tweak the doc build process made by ReadTheDoc.

So I used this mechanism to basically install poetry, export the constraints.txt file at the expected place for docs/requirements.txt, and given that pip supports now editable installs of PEP-517 compliant projects, ... voilà!

tools/release.sh
```diff
@@ -151,6 +142,7 @@ fi
 mkdir kgs
 kgs="kgs/$version"
 pip freeze | tee $kgs
+pip install pytest --constraint ../constraints.txt
```
bmw, line 145: This command fails now because we've changed directories a couple times and `../constraints.txt` is no longer the right path. To fix this, I think we should just install `pytest` with `josepy` above and delete this line.

If after this change the release script works, this PR LGTM!

adferrand: Ok, let's do this. I tried to avoid to modify the behavior of the release script, even if I feel that some logics could may be simplified. Here is a first step.

bmw (approved):
LGTM!

PR #144, opened 2022-10-03
Use proper constraints

bmw:
Our readthedocs build started failing due to a change in pip. See https://readthedocs.org/projects/josepy/builds/18224913/.

Luckily, Poetry added the ability to export constraint.txt files over the weekend. See https://github.com/python-poetry/poetry-plugin-export/blob/dd3b674d21079af615979eda241385f550e3f4ae/CHANGELOG.md#110---2022-10-01.

I updated things to make use of this and you can see the readthedocs build passing with this change at https://readthedocs.org/projects/josepy/builds/18238149/. I also edited the release script to fail fast if poetry export doesn't have constraints.txt support.

Finally, I switched the poetry export flags `--dev` for `--with dev` as `--dev` is deprecated as described at https://github.com/python-poetry/poetry-plugin-export/tree/dd3b674d21079af615979eda241385f550e3f4ae#available-options and was printing deprecation warnings.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,7 +7,7 @@ build:
   jobs:
     pre_install:
       - curl -sSL https://install.python-poetry.org | python3 -
-      - ${HOME}/.local/bin/poetry export -f requirements.txt --dev -E docs --without-hashes > constraints.txt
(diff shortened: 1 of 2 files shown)


alexzorin (approved):
LGTM!

PR #147, opened 2022-10-26
repin dependencies

### LINKED ISSUE #145: Update pyopenssl in poetry.lock
👋 

Over on pyca/cryptography we're trying to ween ourselves off some deprecated OpenSSL APIs (https://github.com/pyca/cryptography/pull/6948). The last blocker is that the josepy tests still use an older pyOpenSSL.

`pyproject.toml` already allows these versions, so they work fine for users, its just a matter of what the tests are run with.

alexzorin:
Fixes #145.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -7,30 +7,22 @@ optional = true
 python-versions = "*"
 
 [[package]]
-name = "atomicwrites"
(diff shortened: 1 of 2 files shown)


bmw (approved):
Thanks Alex.

PR #155, opened 2023-03-01
Set up pre-commit

bmw:
I think this is pretty nice. I've also been considering adding `black` and I'd personally really want this if we start using it.

I ran `poetry lock` to update things because while it shouldn't matter too much, I wanted to match the versions of `flake8` and `isort` used with `pre-commit` and those in `poetry.lock` and the pinned version of `isort` didn't work.


CONTRIBUTING.md
```diff
@@ -50,6 +50,10 @@ $ source .venv/bin/activate
 # (On Windows Powershell)
 $ .\.venv\Script\activate
 ```
+4) Optionally set up [pre-commit](https://pre-commit.com/)
```
alexzorin, line 53: I think a short explanation about what this does would be helpful here.

alexzorin (changes requested):
Works great. That pre-commit project is pretty cool. Just one documentation thing.

bmw:
I also went ahead and set up pre-commit to just test `src` and `test` files for now. Sorry for the additional change, but testing everything is arguably overkill and that approach is likely to cause problems if we take it in the Certbot repo for the reasons being discussed at https://github.com/certbot/certbot/issues/7909.

Thanks so much for all the reviews Alex.

alexzorin (approved):

PR #156, opened 2023-03-06
Update sphinx and stop using pkg_resources

bmw:
https://github.com/certbot/josepy/pull/155 broke our [docs build](https://readthedocs.org/projects/josepy/builds/19657395/) because for some reason [Poetry's solution to dependency resolution in that PR downgraded Sphinx](https://github.com/certbot/josepy/pull/155/files#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89L1050-R1278) to a version that doesn't work on Python 3.10.

This PR initially just fixed that, but then I got errors because [the new version of `setuptools` deprecates `pkg_resources`](https://setuptools.pypa.io/en/latest/history.html#v67-5-0). To fix that, I used the "Attention" message at the top of https://setuptools.pypa.io/en/latest/pkg_resources.html and https://importlib-resources.readthedocs.io/en/latest/index.html. Importing `testdata` as a package like this after adding `__init__.py` Just Works™ for the same reason `import test_util` works.


bmw:
Closing this in favor of #157 and #158.

PR #157, opened 2023-03-07
Fix docs by updating Sphinx

bmw:
https://github.com/certbot/josepy/pull/155 broke our [docs build](https://readthedocs.org/projects/josepy/builds/19657395/) because for some reason Poetry's solution to dependency resolution in that PR downgraded Sphinx to a version that doesn't work on Python 3.10. You can see this downgrade in that PR by looking for the lines under `name = "sphinx"` in `poetry.lock`.

This PR fixes the problem by specifying a requirement on a new version of Sphinx on Python>=3.10.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -168,100 +168,87 @@ files = [
 
 [[package]]
 name = "charset-normalizer"
-version = "3.0.1"
+version = "3.1.0"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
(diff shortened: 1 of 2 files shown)


wgreenberg (approved):

bmw:
Gah. This just [changed the problem](https://readthedocs.org/projects/josepy/builds/19705150/). I'm working on another fix. Thanks for the review though.

bmw:
New PR is #159.

PR #158, opened 2023-03-07
Stop using pkg_resources

bmw:
When working on #157, I got errors because [the new version of `setuptools` deprecates `pkg_resources`](https://setuptools.pypa.io/en/latest/history.html#v67-5-0). To fix that, I used the "Attention" message at the top of https://setuptools.pypa.io/en/latest/pkg_resources.html and https://importlib-resources.readthedocs.io/en/latest/index.html. Importing `testdata` as a package like this after adding `__init__.py` Just Works™ for the same reason `import test_util` works.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,8 +1,9 @@
 """Test utilities."""
+import atexit
+import contextlib
 import os
 from typing import Any
 
-import pkg_resources
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from OpenSSL import crypto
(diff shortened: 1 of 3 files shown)


tests/test_util.py
```diff
@@ -11,17 +13,29 @@
...
+    import importlib_resources
+
+TESTDATA = importlib_resources.files('testdata')
+
 
 def vector_path(*names: str) -> str:
     """Path to a test vector."""
-    return pkg_resources.resource_filename(
-        __name__, os.path.join('testdata', *names))
+    file_manager = contextlib.ExitStack()
```
alexzorin, line 28: It would be nice to have an explanatory comment here.

For me, it was not obvious that `as_file` might generate temporary files on disk that need to be cleaned up.

alexzorin (changes requested):

tests/test_util.py
bmw: That's totally reasonable. I did something slightly different but maybe more useful; I linked to the migration guide I based this code on.

alexzorin (approved):
LGTM!

PR #159, opened 2023-03-07
Actually fix our docs build

bmw:
https://github.com/certbot/josepy/pull/157 fixed the problem locally, but [failed on readthedocs](https://readthedocs.org/projects/josepy/builds/19705150/). I believe that failure is due to https://github.com/python-poetry/poetry-plugin-export/issues/168.

To fix this, I just unconditionally tightened our Sphinx dependency. While I initially tried not to do this, I don't think it's worth the effort and I really doubt it will be a problem for people. If it is, we can always consider doing something fancier in a followup release. I also changed the minimum version to `4.3.0` based on [Sphinx's changelog](https://github.com/sphinx-doc/sphinx/blob/d50ae3c1c917cfa075125ab44c0c8f0eb7aad3ed/CHANGES#L775-L781).

Unfortunately, that change wasn't enough though because `poetry` then selected an ancient version of `flake8` that [fails on Python 3.7](https://github.com/certbot/josepy/actions/runs/4358830770/jobs/7619901487). I fixed this by also specifying a minimum version of that package. The need to keep tracking down minimum versions like this feels somewhat annoying to me, however, in poetry's defense, each solution it presented was valid from its perspective with the dependencies we had specified (until now). I don't think the tigher `flake8` dependency will be a problem since it's just used for testing.

Finally, I tested all of this actually on readthedocs and it passed. See https://readthedocs.org/projects/josepy/builds/19706627/.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -390,14 +390,14 @@ files = [
 
 [[package]]
 name = "docutils"
-version = "0.18.1"
+version = "0.17.1"
 description = "Docutils -- Python Documentation Utilities"
 category = "main"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 files = [
-    {file = "docutils-0.18.1-py2.py3-none-any.whl", hash = "sha256:23010f129180089fbcd3bc08cfefccb3b890b0050e1ca00c867036e9d161b98c"},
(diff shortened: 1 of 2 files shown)


wgreenberg:
ack, good catch! also thanks for the writeup, i was pretty mystified by the error poetry gave.

wgreenberg (approved):

PR #160, opened 2023-03-09
Is this the part where I say "paint it black"?

bmw:
This PR sets up [black](https://github.com/psf/black), the "uncompromising code formatter", which is maintained on GitHub under the Python Software Foundation, has become very popular, and [is used by many projects](https://github.com/psf/black#used-by) including pyca's cryptography which I feel I often point to as an example of a well maintained project. Using a code formatter like this gives our code a consistent style and automatically fixes many problems our linter would otherwise yell about and make us fix manually.

The first commit here was entirely programmatically generated. To create it, first I installed black with `pip` into the virtual environment and then I ran the command put in the commit message.

The second commit was made manually. I added the first commit to `.git-blame-ignore-revs` to avoid masking `git blame` for simple things like `black` changing the quote style. Doing this is [suggested by black](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html) and Alex at https://github.com/certbot/certbot/issues/8818. It's worth noting that GitHub does now support this though as described at https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view. Because of the changes to `.git-blame-ignore-revs`, **this PR should not be squashed when it is merged to preserve the commit hash.**

The second commit also included me rewriting some ignore comments for other linters that `black` didn't understand, but this otherwise was pretty easy and straightforward for me to do.

### DIFF
--- a/tests/jwa_test.py
+++ b/tests/jwa_test.py
@@ -186,12 +186,18 @@ class JWAECTest(unittest.TestCase):
 
         key = JWK.from_json(
             {
-                "d": "Af9KP6DqLRbtit6NS_LRIaCP_-NdC5l5R2ugbILdfpv6dS9R4wUPNxiGw"
-                "-vVWumA56Yo1oBnEm8ZdR4W-u1lPHw5",
-                "x": "AD4i4STyJ07iZJkHkpKEOuICpn6IHknzwAlrf-1w1a5dqOsRe30EECSN4vFxae"
-                "AmtdBSCKBwCq7h1q4bPgMrMUvF",
-                "y": "AHAlXxrabjcx_yBxGObnm_DkEQMJK1E69OHY3x3VxF5VXoKc93CG4GLoaPvphZQv"
-                "Znt5EfExQoPktwOMIVhBHaFR",
+                "d": (
+                    "Af9KP6DqLRbtit6NS_LRIaCP_-NdC5l5R2ugbILdfpv6dS9R4wUPNxiGw"
+                    "-vVWumA56Yo1oBnEm8ZdR4W-u1lPHw5"
+                ),
+                "x": (
+                    "AD4i4STyJ07iZJkHkpKEOuICpn6IHknzwAlrf-1w1a5dqOsRe30EECSN4vFxae"
+                    "AmtdBSCKBwCq7h1q4bPgMrMUvF"
+                ),
+                "y": (
+                    "AHAlXxrabjcx_yBxGObnm_DkEQMJK1E69OHY3x3VxF5VXoKc93CG4GLoaPvphZQv"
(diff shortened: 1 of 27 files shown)


tests/jwa_test.py
```diff
@@ -167,30 +175,33 @@ def test_verify_new_api(self) -> None:
...
-                'd': 'Af9KP6DqLRbtit6NS_LRIaCP_-NdC5l5R2ugbILdfpv6dS9R4wUPNxiGw'
-                     '-vVWumA56Yo1oBnEm8ZdR4W-u1lPHw5',
-                'x': 'AD4i4STyJ07iZJkHkpKEOuICpn6IHknzwAlrf-1w1a5dqOsRe30EECSN4vFxae'
-                     'AmtdBSCKBwCq7h1q4bPgMrMUvF',
-                'y': 'AHAlXxrabjcx_yBxGObnm_DkEQMJK1E69OHY3x3VxF5VXoKc93CG4GLoaPvphZQv'
-                     'Znt5EfExQoPktwOMIVhBHaFR',
-                'crv': 'P-521',
-                'kty': 'EC'
-            })
+                "d": "Af9KP6DqLRbtit6NS_LRIaCP_-NdC5l5R2ugbILdfpv6dS9R4wUPNxiGw"
```
ohemorange, line 189: ok this one I disagree with.

bmw:
I rewrote [my original post/PR description](https://github.com/certbot/josepy/pull/160#issue-1617628680) to better explain what's going on here.

To comment generally on the changes black made though, not all of them are not my personal stylistic preference either. To grab some relevant quotes from [black's README](https://github.com/psf/black#readme):

* "Blackened code looks the same regardless of the project you're reading. Formatting becomes transparent after a while and you can focus on the content instead."
* "Black is opinionated so you don't have to be."

I'm personally convinced and would like to try it. The style it uses is also explained in detail at https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html if you're curious.

tests/jwa_test.py
ohemorange: ok reading black's documentation, I bet it would look nice again if each value was enclosed in parens.

bmw: How does that look for you?

I can also rebase and squash commits if you want since we don't want to let GitHub squash this, but I'm personally not sure it's worth worrying about.

ohemorange: Much better, delighted to see that this worked as expected! Nah no need to rebase.

ohemorange (approved):
We're going to be so stylish.

bmw:
> We're going to be so stylish.

Yes, but if I do the same thing on the Certbot repo after playing with it here, it's going to be much easier to get our tests to pass lint checks again.

PR #161, opened 2023-04-05
Remove setup.cfg

bmw:
We don't need this file and can configure flake8 in tox.ini. See https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations.


alexzorin (approved):
LGTM

PR #163, opened 2023-04-09
Rely on pre-commit

bmw:
I've seen this approach more and more lately — instead of installing tools in both the developer environment and pre-commit where they can have different behavior and be on different versions, just rely on pre-commit. This approach is taken by both https://github.com/tox-dev/tox and https://github.com/pre-commit/pre-commit for example.

I personally like it and think it's a clever way to structure things.


tox.ini
```diff
@@ -6,30 +6,15 @@ max-line-length = 100
 
 [tox]
 isolated_build = true
-envlist = py,black,flake8,isort,mypy
+envlist = mypy,pre-commit,py
```
alexzorin, line 9: What if we added the `Unrecognized environment name` functionality here as well? I am noticing that something like `tox -e black` succeeds, but of course it just runs the unit tests.

alexzorin (changes requested):
One thing I noticed is that removing `black` from the user means that I would have to install `black` separately in order to have VS Code automatically format on save. This is my preferred setup in projects where autoformatting is a thing. On the other hand, I can install it by hand, so it's not a dealbreaker.

tox.ini
bmw: Sure!

bmw:
As for your vscode setup, would you prefer if we added a pinned version of `black` to `pyproject.toml` with a comment that it should be kept in sync with the version in `.pre-commit-config.yaml` as well as a similar comment on the black version in `.pre-commit-config.yaml`? If doing that works nicer for you, I'm fine with that approach.

Thanks so much for all the reviews Alex!

alexzorin:
I'm happy to do that as a personal customization for now, don't think the occasion calls for it yet.

alexzorin (approved):

PR #165, opened 2023-07-10
Deprecate Python 3.7

bmw:
Based on https://github.com/certbot/josepy/pull/132.

Tests may be overkill, but adding some quick tests didn't seem like a bad idea to me.


ohemorange (approved):
lgtm!

PR #182, opened 2024-08-27
Refactoring to drop deprecated usages in `pyOpenSSL` & drop Python 3.7 support

adferrand:
Fixes #181 

**Caution**: This PR introduces backward incompatible changes to the library. A major version of `josepy` should be issued when this PR is merged.
**Caution**: Defensive PR on Certbot should be merged first -> https://github.com/certbot/certbot/pull/9993

This PR removes the APIs deprecated by `pyopenssl` in version 24.2+.

As a consequence, several changes of `josepy` public API are made:
* `josepy.ComparableX509` is removed since the underlying wrapped object is now objects from `cryptography` API (`Certificate` or `CertificateSigningRequest`) and these objects support equality checks.
* methods `encode_csr`, `decode_csr`, `encode_cert` and `decode_cert` now manipulates objects from `cryptography` API
* underlying tests are updated.

Aside these changes, the support of Python 3.7 is dropped, given this PR will require a new major version anyway.

PR #186, opened 2024-09-17
Drop Python 3.7 support and update dependencies

bmw:
We deprecated Python 3.7 support in https://github.com/certbot/josepy/pull/165. This removes that and then I ran `poetry lock` to update dependencies fixing all dependabot alerts.

As pointed out by Alexis at https://github.com/certbot/josepy/pull/182#issuecomment-2333203143, Python 3.7 needs to be dropped to fix security alerts.


ohemorange (approved):

bmw:
it's not a big deal in my opinion, but this PR missed a few references to python 3.7. i cleaned them up in https://github.com/certbot/josepy/pull/195

PR #187, opened 2024-09-17
Fix CI

bmw:
I accidentally broke CI in https://github.com/certbot/josepy/pull/160. In that PR I tried to simplify our CI setup by just running `tox` which normally runs all standard tests, but I missed that `TOXENV` was set at the top of the file causing `tox` with no other arguments to only run `mypy`!

This PR fixes that, updates `.pre-commit-config.yaml` so it works with modern Pythons, and applies the minor changes from `black`.


wgreenberg (approved):
:+1:

PR #194, opened 2024-12-18
deprecate python 3.8 support

bmw:
this is based on https://github.com/certbot/josepy/pull/165


bmw:
sorry for the followup commits here. i initially started removing `importlib_resources` as it's not needed once we only depend on python 3.9+, but since this PR only deprecates python 3.8 support, we can't do that quite yet

ohemorange (approved):
lgtm!

PR #200, opened 2025-01-03
Filter warning caused by deprecated pyopenssl usage of datetime

ohemorange:
Since we're likely to deprecate this whole function soon anyway, I am putting the warning filter directly in the code. This fixes the tests currently [failing](https://github.com/certbot/josepy/actions/runs/12604702846) on master.


wgreenberg (approved):
:+1:

PR #204, opened 2025-01-15
Deprecate usage of `ComparableX509` in encoding and decoding functions

ohemorange:
Largely based on https://github.com/certbot/josepy/pull/202.

Marks `json_util.decode_cert` and `json_util.decode_csr` as deprecated, and creates `json_util.decode_cert_cryptography` and `json_util.decode_csr_cryptography` in their place.

Deprecates accepting `ComparableX509` in `json_util.encode_cert` and `json_util.encode_csr`, and accepts the relevant `cryptography.x509` objects in their place (either is accepted as input for now).


ohemorange:
Sorry wrong PR

PR #214, opened 2025-01-31
drop python 3.8 support

bmw:
you can compare this to the PR that did this for 3.7 which is https://github.com/certbot/josepy/pull/186


ohemorange (approved):
lgtm!

PR #224, opened 2025-05-28
Add support for Python 3.14

cdce8p:
With [PEP 649](https://peps.python.org/pep-0649/) annotation handling in Python 3.14 will change. The [Python docs](https://docs.python.org/3.14/library/annotationlib.html#using-annotations-in-a-metaclass) provide a good guide on how to access all annotations in 3.14 and beyond. I've updated the code in `json_util.py` to use the new functions.

Furthermore `argparse.FileType` will be deprecated. It's recommend to use the regular `open` contextmanager instead.

Lastly, I've added `3.14` to the CI test matrix and the corresponding classifier.

--
If accepted, it would be awesome if a new version could be released soon as this would unblock further testing with Python 3.14 in downstream projects.


wgreenberg (approved):
apologies for the late review, but this is great, thanks! tragically there's now a conflict in the changelog, but after that's resolved, this is good to go

wgreenberg (approved):

pyproject.toml
```diff
@@ -79,7 +80,7 @@ jws = "josepy.jws:CLI.run"
 line-length = 100
 # This should be kept in sync with the version of Python specified in poetry's
 # dependencies above.
-target-version = ['py39', 'py310', 'py311', 'py312', 'py313']
+target-version = ['py39', 'py310', 'py311', 'py312', 'py313'] # TODO add 'py314'
```
ohemorange, line 83: TODOs should have an associated issue tracking the plan to complete them

src/josepy/json_util.py
```diff
@@ -233,7 +237,17 @@ def __new__(
...
                     # Error out if it is not the case.
-                    if key not in namespace.get("__annotations__", {}):
+                    if "__annotations__" in namespace:
+                        annotations = namespace["__annotations__"]
+                    elif sys.version_info >= (3, 14) and (
+                        annotate := annotationlib.get_annotate_from_class_namespace(namespace)
+                    ):
+                        annotations = annotationlib.call_annotate_function(
+                            annotate, format=annotationlib.Format.FORWARDREF
+                        )
```
ohemorange, line 247: I personally think use of walrus operators should be limited, and certainly if statements should not be complex enough to require multiple lines. What about the following:

```suggestion
                    elif sys.version_info >= (3, 14):
                        annotate = annotationlib.get_annotate_from_class_namespace(namespace)
                        if annotate: 
                            annotations = annotationlib.call_annotate_function(
                                annotate, format=annotationlib.Format.FORWARDREF
                            )
                        else:
                            annotations = {}
```

This seems more readable to me, but perhaps I'm being curmudgeonly about new language features here.

.github/workflows/check.yaml
```diff
@@ -38,12 +39,17 @@ jobs:
   # This job runs our tests like external parties such as packagers.
   external:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
```
ohemorange, line 45: I'm not sure it's necessary to test as external packagers do for every version of python. This test is mainly to make sure we're not accidentally on things in `dev-extras` or the like. What is the goal of adding five more tests here?

ohemorange (changes requested):

pyproject.toml
cdce8p: Not sure that's really necessary here. I only added the comment since the current black version doesn't support `py314`. The line will also become redundant once the build-system is updated to poetry `v2` which supports `project.requires-python`.

src/josepy/json_util.py
cdce8p: Not sure that's an improvement. To fully capture all options, you'd need to `else` clauses. So overall the `:=` use makes it more readable IMO. `:=` is also used in the example from the Python docs I've adapted this section from.
https://docs.python.org/3.14/library/annotationlib.html#using-annotations-in-a-metaclass

```py
if "__annotations__" in namespace:
    annotations = namespace["__annotations__"]
elif sys.version_info >= (3, 14):
    annotate = annotationlib.get_annotate_from_class_namespace(namespace)
    if annotate: 
        annotations = annotationlib.call_annotate_function(
            annotate, format=annotationlib.Format.FORWARDREF
        )
    else:
        annotations = {}
else:
    annotations = {}
```

.github/workflows/check.yaml
ohemorange: The warning is just saying that it's falling back to a default, which is fine. But even if it makes sense to run tests on `pip install`ed josepy for all python versions, this PR is not the place to do it. If the warning is an issue, this change will keep things as they were before, but without the warning. In which case we also do not need `allow-prereleases: true` added below.

```suggestion
        python: ["3.13"]
```

pyproject.toml
ohemorange: Those are the sorts of details that would be great in an issue so that we remember to update this code when they become available.

src/josepy/json_util.py
ohemorange: Like list comprehensions, walrus operators should only be used in simple, readable cases. This case differs from the python example because the original `elif` is checking multiple things, creating an overly long expression to check. I think that code block directly above looks great, as it's what I suggested above. It makes it very clear what happens in the various cases, and in the future when only 3.14+ is available, it will be very neat and clear to update.

If you're really excited about walruses and don't like else statements, the following is also tolerable:

```py
annotations = {}
if "__annotations__" in namespace:
    annotations = namespace["__annotations__"]
elif sys.version_info >= (3, 14):
    if annotate := annotationlib.get_annotate_from_class_namespace(namespace):
        annotations = annotationlib.call_annotate_function(
            annotate, format=annotationlib.Format.FORWARDREF
        )
```

or, without the walrus:

```py
annotations = {}
if "__annotations__" in namespace:
    annotations = namespace["__annotations__"]
elif sys.version_info >= (3, 14):
    annotate = annotationlib.get_annotate_from_class_namespace(namespace)
    if annotate: 
        annotations = annotationlib.call_annotate_function(
            annotate, format=annotationlib.Format.FORWARDREF
        )
```

My main issue here is the length and complexity of the expression to be evaluated in the `elif`. Solutions that avoid that problem work for me.

ohemorange (changes requested):
> However my main focus is to add support for Python 3.14 and not to insist on my coding style.

It is important to have a readable, maintainable codebase. Coding style is in fact a part of reviewing PRs.

pyproject.toml
cdce8p: Opened #232 and updated the comment here.

ohemorange (approved):
Thank you! lgtm!

ohemorange:
We plan to do a new release soon!

PR #240, opened 2025-09-12
deprecate python 3.9 support

bmw:
fixes https://github.com/certbot/josepy/issues/239. this PR was based on https://github.com/certbot/josepy/pull/194

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,9 @@ Changelog
 2.2.0 (main)
 ------------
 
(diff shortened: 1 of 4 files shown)


bmw:
i don't think so. grepping the code for various versions of references to python 3.9 and 3.10, i'm not seeing anything else that should be done when we drop python 3.9 support like there was when [we dropped python 3.8](https://github.com/certbot/josepy/pull/214)

if you think i'm missing something tho, plz lmk!

PR #266, opened 2026-04-03
drop py39 support

### LINKED ISSUE #245: drop python 3.9 support
now that josepy 2.2.0 is out deprecating python 3.9 support, we can do this

the PR that did this for python 3.8 can be seen at https://github.com/certbot/josepy/pull/214

bmw:
fixes https://github.com/certbot/josepy/issues/245

you can compare this to the PR that did this for 3.8 which is https://github.com/certbot/josepy/pull/214

i don't think this PR needs two reviews


ohemorange (approved):
lgtm

PR #276, opened 2026-07-04
Replace 512-bit DSA test key with a 1024-bit key

reaperhulk:
An upcoming pyca/cryptography change will reject 512-bit DSA keys. Regenerate the test vector at 1024 bits, rename it to dsa1024_key.pem, and update the test and README references.


wgreenberg (approved):
good looking out, thanks!

ohemorange (approved):

